PR #282, opened 2020-03-05
refactor(selected-layers): add selected layer ids to url

KatvonRivia:


src/scripts/libs/globe-url-parameter.ts
```diff
@@ -19,8 +21,8 @@ export function parseUrl(): GlobeState | null {
   }
 
   const splitted = globeParam.split(char);
-
-  if (splitted.length !== 8) {
+  console.log(splitted);
```
pwambach: remove `console.log`

PR #289, opened 2020-03-10
refactor(router): add routes, add custom hook to get story params

KatvonRivia:


src/scripts/components/story/story.styl
```diff
@@ -2,23 +2,9 @@
 
 .story
   display: flex
-  flex-shrink: 0
-  flex-direction: column
-  width: $storyWidth
+  flex-direction: row
+  width: 100%
```
pwambach: Do we need `width` here?

PR #291, opened 2020-03-12
refactor(story): add story header and content to all story modes

KatvonRivia:


src/scripts/components/stories-selector/stories-selector.styl
```diff
@@ -11,18 +13,20 @@
 .header
   display: flex
   justify-content: space-between
-  padding: 10px
+  align-items: center
+  height: emCalc(72px)
```
pwambach: As we use this value on multiple elements we could put it in the variables file

src/scripts/components/story-header/story-header.styl
```diff
@@ -0,0 +1,23 @@
+@require '../../../variables.styl'
+
+.storyHeader
+  display: flex
+  flex-shrink: 0
+  justify-content: space-between
+  align-items: center
+  height: emCalc(72px)
```
pwambach: Put in variables file, or do you still plan to make a general header component? Would probably be the best solution

src/scripts/components/story-header/story-header.tsx
```diff
@@ -0,0 +1,33 @@
...
+import styles from './story-header.styl';
+
+interface Props {
+  mode: StoryMode;
+  storyIds: string[];
+}
+
+const StoryHeader: FunctionComponent<Props> = ({mode, storyIds}) => {
+  const storyList = useSelector(storyListSelector);
+  const storyId = storyIds[0];
```
pwambach: Why is the index always `0`? And don't we have the currently selected story already in the story component? Can't we just pass the title to the story header component?

src/scripts/components/story-list/story-list.styl
```diff
@@ -1,9 +1,14 @@
 @require '../../../variables.styl'
 
 .storyList
-  display: grid
-  padding: 20px
+  overflow-y: scroll
```
pwambach: `.storyList` and `.storyListGrid` have `overflow-y: scroll`. Are you sure we need both divs/classes?

src/scripts/components/story/story.tsx
```diff
@@ -1,37 +1,61 @@
...
     <div className={storyClasses}>
-      {/* <StoryHeader /> */}
-      <StoryContent />
-      <Globes />
-      <StoryMedia />
-      {/* <StoryFooter /> */}
+      <StoryHeader mode={mode} storyIds={storyIds} />
+      <main className={styles.main}>
+        {/* Instead of rendering only the currect slide we map over all slides to
+        enforce a newly mounted component when the pageNumber changes */}
```
pwambach: Please use `slideNumber` instead of `pageNumber`

src/scripts/components/story/story.tsx
```diff
@@ -36,16 +31,21 @@ const Story: FunctionComponent = () => {
     currentStoryId && dispatch(fetchStory(currentStoryId));
   }, [dispatch, currentStoryId]);
 
+  const storyList = useSelector(storyListSelector);
+  const currentStory = storyList.find(story => story.id === currentStoryId);
```
pwambach: Hmm now we have `currentStory` and `selectedStory`. Can we move these two lines also in the storyParams hook and just return a new prop storyListItem?

PR #297, opened 2020-03-17
refactor(header): add header component

KatvonRivia:


src/scripts/components/header/header.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <div className={headerClasses}>
+      <Button
+        className={styles.backButton}
+        icon={ArrowBackIcon}
+        label={backButtonId}
+        link={backLink}
+      />
+      <h1 className={styles.title}>{title}</h1>
+      {selectedIds && (
```
pwambach: Make this dynamic by using the children property here

PR #300, opened 2020-03-18
refactor(story-content): add story gallery and videos

KatvonRivia:


src/scripts/components/story-content/story-content.styl
```diff
@@ -1,19 +1,24 @@
 @require '../../../variables.styl'
 
 .content
-  padding: emCalc(80px) emCalc(56px)
-  color: $textWhite
+  overflow-y: scroll
```
pwambach: `overflow-y: auto` please, otherwise i get these ugly empty scrollbars on my machine with a mouse

src/scripts/components/story/story.tsx
```diff
@@ -40,6 +42,16 @@ const Story: FunctionComponent = () => {
...
   }
 
+  const getRightSideComponent = (slide: Slide, story: StoryType) => {
+    if (slide.images) {
+      return <StoryMedia images={slide.images} storyId={story.id} />;
+    } else if (slide.videoId) {
+      return <StoryVideo videoId={slide.videoId} />;
+    }
+
+    return <Globes />;
```
pwambach: hmm I don't think the globe transitions will work this way, could be that they jump to the next position instead of a flyto but let's see...

PR #308, opened 2020-03-24
refactor(layer-selector): add remove button to selected main layer

KatvonRivia:


src/scripts/components/selected-layer-list-item/selected-layer-list-item.styl
```diff
@@ -21,6 +21,7 @@
   border-radius: 50%
 
 .removeIcon
+  display: contents
```
KatvonRivia: What do you think of this?

Without it, the remove button ist not centered in the div, even though all items are center aligned: 
image

with display: contents:
image

pwambach: hmm support for this seems to be ok but not really good. What's the problem with aligning it with display: flex or something similiar? I would prefer the "normal" way

KatvonRivia: Okay 👍

PR #312, opened 2020-03-25
refactor(story): add shortText, change styles

### LINKED ISSUE #309: Narratives check example story
Here is the content for the example story:
[Story Content](https://drive.google.com/open?id=16HDxb8a-a4ZpLM5V7EbrpWUXT46sv2d_) (has probably not changed)
[Exhibition version](https://drive.google.com/open?id=1EhIwiSdFJzSF3iZKa7d_ScyrE7KlKg6K)

- [ ] slide 4 should be an interactive compare between two datasets
- [ ] inline images (see Figma for updated design)
  - center image horizontally
  - we might use the alt text of the image as image caption?!
image
- the final datasets are the ones PW is working on right now. Use placeholder data for now.
- [ ] images in main area should be displayed completely (rather object-fit contain than cover)
- [ ] New Video for slide 3:
  http://www.planetaryvisions.com/download.php?id=53
  sst19_nocaptions.mp4
  sst19_nocaptions.srt

KatvonRivia:
closes #309


src/scripts/components/story-content/story-content.styl
```diff
@@ -15,10 +16,18 @@
     font-family: NotesEsa
     line-height: emCalc(16px)
 
-  img
-    width: 50%
+  img, em
```
pwambach: this is a bit confusing, most of the properties have nothing to do with the img. could you split these up, so that both together only get the properties they need. Like
```
img, em
  ...
img
  ...
em
  ...
```

src/scripts/components/story-pagination/story-pagination.tsx
```diff
@@ -51,8 +52,14 @@ const StoryPagination: FunctionComponent<Props> = ({
           nextSlideLink && history.push(nextSlideLink);
         }
       }
+      if (isShowcaseMode) {
```
pwambach: use `else if (event.keyCode === 27)` here to get rid off the complexity

PR #313, opened 2020-03-25
refactor(header): align header items

KatvonRivia:


src/scripts/components/header/header.tsx
```diff
@@ -25,6 +25,7 @@ const Header: FunctionComponent<Props> = ({
       link={backLink}
     />
     <h1 className={styles.title}>{title}</h1>
+    {!children && <div className={styles.empty}></div>}
```
pwambach: Probably better to wrap the children in a div and give the wrapper div these styles. This way we don't need the empty div and the showcase selector does not need any styles specific to the header

src/scripts/components/showcase-selector/showcase-selector.tsx
```diff
@@ -32,7 +32,7 @@ const ShowcaseSelector: FunctionComponent = () => {
         backLink="/"
         backButtonId="backToDataMode"
         title={intl.formatMessage({id: 'showcaseMode'})}>
-        <div className={styles.playButton}>
+        <React.Fragment>
```
pwambach: Do we need this here and why? if we really need a div for styling issues then it should be a real div and not a fragment.

src/scripts/components/header/header.styl
```diff
@@ -24,5 +24,14 @@
   white-space: nowrap
   font-size: emCalc(18px)
 
-.empty
+.playButton
```
pwambach: Please use a more general name here. The whole purpose of the children here is that it doesn't matter what you render inside, so it could also be something other then a play button

src/scripts/components/header/header.styl
```diff
@@ -24,7 +24,7 @@
   white-space: nowrap
   font-size: emCalc(18px)
 
-.playButton
+.controls
```
pwambach: Sorry but please use a more general name like `rightContent` or something. This could be **anything** not only controls

PR #318, opened 2020-03-27
refactor(slider): implement new slider design

### LINKED ISSUE #319: Timeslider

KatvonRivia:
closes #319

### DIFF
--- a/src/scripts/components/data-set-info/data-set-info.styl
+++ b/src/scripts/components/data-set-info/data-set-info.styl
@@ -2,7 +2,7 @@
 
 .dataSetInfo
   position: absolute
-  bottom: emCalc(24px, 16px)
+  bottom: emCalc(24px)
   display: flex
   flex-direction: row
   width: 100%
--- a/src/scripts/components/data-set-info/data-set-info.styl
+++ b/src/scripts/components/data-set-info/data-set-info.styl
@@ -2,7 +2,7 @@
 
 .dataSetInfo
   position: absolute
-  bottom: emCalc(14px)
+  bottom: emCalc(24px, 16px)
   display: flex
   flex-direction: row
   width: 100%
@@ -14,13 +14,10 @@
   align-items: center
   width: 100%
 
-  &:hover, &:focus
-    svg
-      fill: $main
-
 .title
   margin: 0
   color: $textWhite
+  font-weight: normal
   font-size: 1em
 
 .buttons
--- a/src/scripts/components/info-button/info-button.tsx
+++ b/src/scripts/components/info-button/info-button.tsx
@@ -22,7 +22,9 @@ const InfoButton: FunctionComponent<Props> = ({layer}) => {
         <Button
           icon={InfoIcon}
           className={styles.infoButton}
-          onClick={() => setShowMenu(true)}
+          onClick={() => {
+            setShowMenu(true);
+          }}
         />
 
         {showMenu && (
--- a/src/scripts/components/info-button/info-button.tsx
+++ b/src/scripts/components/info-button/info-button.tsx
@@ -22,9 +22,7 @@ const InfoButton: FunctionComponent<Props> = ({layer}) => {
         <Button
           icon={InfoIcon}
           className={styles.infoButton}
-          onClick={() => {
-            setShowMenu(true);
-          }}
+          onClick={() => setShowMenu(true)}
         />
 
         {showMenu && (
--- a/src/scripts/components/time-slider/time-slider.styl
+++ b/src/scripts/components/time-slider/time-slider.styl
@@ -67,7 +67,7 @@ compareThumb()
   top: 0
   z-index: 1
   display: flex
-  justify-content: space-between
+  overflow-x: hidden
   height: $range-height
   border-radius: emCalc(2px)
   background: $main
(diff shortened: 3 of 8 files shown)


src/scripts/components/data-set-info/data-set-info.styl
```diff
@@ -2,7 +2,7 @@
 
 .dataSetInfo
   position: absolute
-  bottom: emCalc(14px)
+  bottom: emCalc(24px, 16px)
```
pwambach: Isn't 16 the default value?

src/scripts/components/info-button/info-button.tsx
```diff
@@ -22,7 +22,9 @@ const InfoButton: FunctionComponent<Props> = ({layer}) => {
         <Button
           icon={InfoIcon}
           className={styles.infoButton}
-          onClick={() => setShowMenu(true)}
+          onClick={() => {
+            setShowMenu(true);
```
pwambach: Remove unnecessary `{}` brackets

src/scripts/components/time-slider/time-slider.styl
```diff
@@ -2,42 +2,109 @@
 
 $range-height = 5px
 
+thumb()
+  width: 16px
+  height: 16px
+  border-radius: 50%
+  background: $main
+  box-shadow: 0px 1px 5px rgba(0, 0, 0, 0.2), 0px 3px 4px rgba(0, 0, 0, 0.12), 0px 2px 4px rgba(0, 0, 0, 0.14)
```
pwambach: Please make a separate mixin for the box shadow so that we have this declaration only once

src/scripts/components/time-slider/time-slider.tsx
```diff
@@ -89,39 +92,68 @@ const TimeSlider: FunctionComponent = () => {
     };
   };
 
+  const setTimeOutput = (targetValue: string, min: number, max: number) => {
```
pwambach: If you only want to set the style on an element you can do it with a normal inline style attribute and don't need to use `useRef`

src/scripts/components/time-slider/time-slider.tsx
```diff
@@ -89,39 +92,68 @@ const TimeSlider: FunctionComponent = () => {
...
+    const currentTime = parseInt(targetValue, 10);
+    const outputPosition = Number(((currentTime - min) * 100) / (max - min));
+
+    if (timeOutput.current) {
+      timeOutput.current.style.left = `calc(${outputPosition}% + (${8 -
+        outputPosition * 0.15}px))`;
+      timeOutput.current.style.display = 'flex';
```
pwambach: Please set non-dynamic styles in css

src/scripts/components/time-slider/time-slider.tsx
```diff
@@ -89,39 +92,68 @@ const TimeSlider: FunctionComponent = () => {
...
+          </output>
           {rangeMain && (
-            <div
-              className={styles.rangeMain}
-              style={getRangeStyle(rangeMain.min, rangeMain.max)}></div>
+            <div className={styles.mainTrack}>
+              <div
+                className={styles.rangeMain}
+                style={getRangeStyle(rangeMain.min, rangeMain.max)}>
+                {rangeMain.timestamps.map(timestamp => (
```
pwambach: Please calculate the correct position for every tick based on the timestamp

src/scripts/components/time-slider/time-slider.styl
```diff
@@ -2,42 +2,120 @@
...
+  height: 16px
+  border-radius: 50%
+  background: $main
+  cursor: pointer
+
+compareThumb()
+  width: 10px
+  height: 25px
+  border-radius: emCalc(4px)
+
```
pwambach: No need for the extra new line when using mixins

KatvonRivia: That happens automatically. Maybe it's a lint config.

pwambach: Could you check if we can deactivate it in the stylint settings?

src/scripts/components/time-slider/time-slider.tsx
```diff
@@ -89,39 +90,83 @@ const TimeSlider: FunctionComponent = () => {
...
+          </output>
           {rangeMain && (
-            <div
-              className={styles.rangeMain}
-              style={getRangeStyle(rangeMain.min, rangeMain.max)}></div>
+            <div className={styles.mainTrack}>
+              <div
+                className={styles.rangeMain}
+                style={getRangeStyle(rangeMain.min, rangeMain.max)}>
+                {rangeMain.timestamps.map(timestamp => {
```
pwambach: Please calculate the ticks outside so that the jsx markup is more readable. You can pre-calculate the positions in an array and iterate over it here instead of the timestamps.

src/scripts/components/time-slider/time-slider.tsx
```diff
@@ -89,39 +90,83 @@ const TimeSlider: FunctionComponent = () => {
...
           {rangeCompare && (
-            <div
-              className={styles.rangeCompare}
-              style={getRangeStyle(rangeCompare.min, rangeCompare.max)}></div>
+            <div className={compareStyles}>
+              <div
+                className={styles.rangeCompare}
+                style={getRangeStyle(rangeCompare.min, rangeCompare.max)}>
+                {rangeCompare.timestamps.map(timestamp => {
+                  const tickPosition = Number(
```
pwambach: Same here

src/scripts/components/time-slider/time-slider.tsx
KatvonRivia: I moved the calculation to a function, like we did it for the range style. Is that fine too?

PR #326, opened 2020-04-02
refactor(styles): add new styles and icons

KatvonRivia:


src/scripts/components/globe-navigation/globe-navigation.styl
```diff
@@ -7,9 +7,13 @@
...
 .compassIcon
   padding-right: emCalc(16px)
   padding-left: emCalc(25px)
+
+.compassIcon svg, .downloadIcon svg
+  fill: $textWhite
+
+.compassIcon svg
+  &:hover
+    transform: rotate(-45deg)
```
pwambach, line 19: I thought the compass icon should rotate with the globe?

src/scripts/components/presentation-selector/presentation-selector.tsx
```diff
@@ -16,8 +18,9 @@ const PresentationSelector: FunctionComponent = () => {
       <Header
         backLink="/"
         backButtonId="backToDataMode"
-        title={intl.formatMessage({id: 'presenterMode'})}
-      />
+        title={intl.formatMessage({id: 'presenterMode'})}>
+        <Button icon={ShareIcon} onClick={() => console.log('placeholder')} />
```
pwambach, line 22: Good thing that we made this easy to extend :)

src/scripts/components/showcase-selector/showcase-selector.tsx
```diff
@@ -36,12 +36,19 @@ const ShowcaseSelector: FunctionComponent = () => {
...
           values={{numberSelected: selectedIds.length}}
         />
-        <Button
-          className={styles.button}
-          label={'play'}
-          link={`/showcase/${selectedIds.join('&')}/0/title`}
-          icon={PlayIcon}
-        />
+        {selectedIds.length === 0 ? (
+          <Button
```
pwambach, line 40: The button should have a "disabled" property which sets the real disabled attribute on the button element (if there is one) and handles the styling automatically

src/scripts/components/story/story.tsx
```diff
@@ -78,8 +80,9 @@ const Story: FunctionComponent = () => {
         <Header
           backLink={`/${mode.toString()}`}
           backButtonId="backToStories"
-          title={storyListItem.title}
-        />
+          title={storyListItem.title}>
+          <Button icon={ShareIcon} onClick={() => console.log('placeholder')} />
```
pwambach, line 84: Because it's used multiple times we should make a ShareButton component out of it

src/scripts/components/showcase-selector/showcase-selector.tsx
KatvonRivia: Yeah i thought so too but in the button component we either have a button or a link. And in this case it's a link.

pwambach: yes, but the cool thing about this component is that you shouldn't care when using it. so when you set `disabled` the button should get the disabled attribute and be styles correctly and a link tag should prevent the ink from being executed (preventDefault) and also be styled like it is disabled

src/scripts/components/globe-navigation/globe-navigation.styl
KatvonRivia: You're right. It should move with the globe 👍

src/scripts/components/story/story.tsx
KatvonRivia: Yes makes sense, i will also create a separate PR for that after my vacations. :)

src/scripts/components/button/button.tsx
```diff
@@ -16,13 +18,20 @@ const Button: FunctionComponent<Props> = ({
   label,
   link,
   icon: Icon,
+  disabled,
```
pwambach, line 21: you can set "false" as default value here to be more explicit

src/scripts/components/button/button.tsx
```diff
@@ -16,13 +18,20 @@ const Button: FunctionComponent<Props> = ({
...
   className = '',
   onClick
 }) => {
-  const classes = `${styles.button} ${className}`;
+  const classes = cx(
+    `${styles.button} ${className}`,
```
pwambach, line 26: with the cx function you can write this as
```
cx(
    styles.button,
    className,
    disabled && styles.disabled
    disabled && className
)
```

PR #327, opened 2020-04-02
feat(story-footer): slide in story footer on mouse move

### LINKED ISSUE #316: Presenter/showcase mode: hide header and footer
- only  show  header and footer on hover
- hide header and footer after a short  timeout

KatvonRivia:
closes #316

### DIFF
--- a/src/scripts/components/story-footer/story-footer.styl
+++ b/src/scripts/components/story-footer/story-footer.styl
@@ -6,3 +6,10 @@
   height: emCalc(72px)
   height: $footerHeight
   background-color: $darkGrey1
+
+.fadeOutFooter
+  opacity: 0
+  transition: opacity 0.4s ease-in-out
+
+  &:hover
+    opacity: 1
--- a/src/scripts/components/story-footer/story-footer.styl
+++ b/src/scripts/components/story-footer/story-footer.styl
@@ -6,10 +6,8 @@
   height: emCalc(72px)
   height: $footerHeight
   background-color: $darkGrey1
+  transition: transform 0.6s ease-in-out
+  transform: translateY(0%)
 
-.fadeOutFooter
-  opacity: 0
-  transition: opacity 0.4s ease-in-out
-
-  &:hover
-    opacity: 1
+.slideOutFooter
+  transform: translateY(100%)
--- a/src/scripts/hooks/use-mouse-move.ts
+++ b/src/scripts/hooks/use-mouse-move.ts
@@ -5,11 +5,16 @@ export const useMouseMove = () => {
 
   useEffect(() => {
     let timer = 0;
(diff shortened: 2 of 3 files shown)


src/scripts/components/story-footer/story-footer.styl
```diff
@@ -6,3 +6,10 @@
   height: emCalc(72px)
   height: $footerHeight
   background-color: $darkGrey1
+
+.fadeOutFooter
+  opacity: 0
+  transition: opacity 0.4s ease-in-out
+
+  &:hover
```
pwambach, line 14: I'm not sure if it should work this way. In my understanding the header and footer should only be visible when the mouse has moved and disappear after a few seconds when no more interactions happened. Like in the Earth View extension.
Also I thought we they should slide in/out instead of fading. Could you check that gain with Patrick please.

KatvonRivia: Talked to him, he said that showing the items when mouse is moving would be cool. 👍

src/scripts/hooks/use-mouse-move.ts
```diff
@@ -0,0 +1,15 @@
+import {useEffect, useState} from 'react';
+
+export const useMouseMove = () => {
+  const [mouseMove, setMouseMove] = useState(true);
+
+  useEffect(() => {
+    let timer = 0;
+    window.onmousemove = () => {
```
pwambach, line 8: use window.addEventListener instead of onmousemove and removeEventListener in the clean up function. Using onmousemove function can cause errors if anything else relies on this function. addEventlistener is the more general solution

pwambach: Also the cleanup method should clear the timeout too

src/scripts/hooks/use-mouse-move.ts
```diff
@@ -0,0 +1,20 @@
...
+  const [mouseMove, setMouseMove] = useState(true);
+
+  useEffect(() => {
+    let timer = 0;
+    const handleMouseMove = () => {
+      setMouseMove(true);
+      timer = window.setTimeout(() => setMouseMove(false), 5000);
+    };
+    window.addEventListener('mousemove', handleMouseMove);
+    const cleanUp = () => {
```
pwambach, line 13: add a new line and return the function directly please

src/scripts/hooks/use-mouse-move.ts
```diff
@@ -0,0 +1,20 @@
...
+      window.removeEventListener('mousemove', handleMouseMove);
+      clearTimeout(timer);
+    };
+    return cleanUp;
+  }, []);
+  return mouseMove;
```
pwambach, line 19: also a new line before return

PR #328, opened 2020-04-02
refactor(menu): implement new menu design

KatvonRivia:


src/scripts/components/menu/menu.tsx
```diff
@@ -17,50 +16,107 @@ import styles from './menu.styl';
...
     {
-      id: 'presenterMode',
-      link: '/present',
-      icon: PresenterIcon
-    },
-    {
-      id: 'showcaseMode',
-      link: '/showcase',
-      icon: ShowCaseIcon
+      id: 'modes',
```
KatvonRivia, line 19: This seems too long now. Where is a better space to have it?

src/scripts/components/overlay/overlay.styl
```diff
@@ -14,11 +14,24 @@
...
   width: 60%
-  height: 80%
+  height: 70%
 
 .closeButton
   position: absolute
   top: emCalc(24px)
   right: emCalc(32px)
+
+.logo
```
pwambach, line 28: Should the logo be always on every overlay? My first thought about this was that it should be implemented in the specific overlays content

PR #336, opened 2020-04-17
feat(compass): sync compass with globe

KatvonRivia:


src/scripts/components/globe-navigation/globe-navigation.styl
```diff
@@ -7,13 +7,18 @@
...
+.compass
+  margin-right: emCalc(16px)
+  margin-left: emCalc(25px)
+  width: 24px
+  height: 24px
 
-.compassIcon svg, .downloadIcon svg
-  fill: $textWhite
+.compass svg
+  fill: #202B33
```
pwambach, line 17: What color is thta? Wouldn't it be better if the icon defines the default fill color itself and we overwrite only on hover?

src/scripts/components/globe-navigation/globe-navigation.styl
```diff
@@ -7,13 +7,18 @@
   z-index: 1
   display: flex
 
-.compassIcon
-  padding-right: emCalc(16px)
-  padding-left: emCalc(25px)
+.compass
+  margin-right: emCalc(16px)
+  margin-left: emCalc(25px)
```
pwambach, line 12: Why do we need these strange margins? Can we solve this in a better way with flexbox?

src/scripts/components/globe-navigation/globe-navigation.tsx
```diff
@@ -17,6 +18,7 @@ const GlobeNavigation: FunctionComponent = () => {
   const dispatch = useDispatch();
   const defaultView = config.globe.view;
   const projectionState = useSelector(projectionSelector);
+  const globecheck = useSelector(globeViewSelector);
```
pwambach, line 21: If we use the position from the redux store we only get a new value when the globe movement ends. It would be cooler if the icon rotation updates in realtime. Perhaps we can do a little trick and set the rotation as a css variable in the globe component.

src/scripts/components/globe-navigation/globe-navigation.tsx
```diff
@@ -32,11 +34,12 @@ const GlobeNavigation: FunctionComponent = () => {
...
     <div className={styles.globeNavigation}>
       <Button label={label} onClick={() => onProjectionHandler()} />
-      <Button
-        className={styles.compassIcon}
-        icon={CompassIcon}
+      <div
+        className={styles.compass}
         onClick={() => dispatch(setFlyToAction({...defaultView}))}
-      />
+        style={{transform: `rotate(${globecheck.position.longitude}deg)`}}>
```
pwambach, line 40: Using the longitude seems wrong because the north arrow of the icon rotates even if the globe's north is up. One of the camera rotation values could work better

PR #342, opened 2020-05-07
docs(data-pipeline): add notes file

pwambach:


pwambach:
Added this as a wiki page. Seemed liked a better place for these kind of infos. Will close this PR...

PR #381, opened 2020-07-13
chore(basemaps): add script to manually update basemaps on storage

pwambach:


scripts/update-layer-basemap.sh
```diff
@@ -0,0 +1,11 @@
+# Updates the basemap property in a layers metadata.json
+# Usage: ./scripts/updae-layer-basemap.sh <layerId> <newBasemap>
+
+id=$1
+basemap=$2
+path=gs://esa-cfs-tiles/0.6.1/$id/metadata.json
```
pmast, line 6: Should we take the version from the Environment? Or as parameter?

pwambach: yes, makes sense

PR #387, opened 2020-07-15
chore(metadata): script to update all metadata

pmast:


scripts/update-metadata.sh
```diff
@@ -0,0 +1,57 @@
...
+layersConfig=$(cat $layersConfigFile)
+#keys=$(jq -r 'keys[]' <<< "$layersConfig" | head -n 1)
+
+# take optional layer ID from command line argument
+layerID=$1 
+
+# check if layer ID is set
+if [ -z $layerID ]
+then
+    keys=$(jq -r 'keys[]' <<< "$layersConfig")
```
pwambach, line 19: can't we just say `keys=$(jq -r 'keys[]' $layersConfigFile)` here?

scripts/update-metadata.sh
```diff
@@ -0,0 +1,57 @@
...
+    echo $datasetId 
+
+    # download metadata file
+    path=gs://esa-cfs-tiles/$version/$datasetId/metadata.json
+    metadataFileName=$datasetId-metadata.json
+    gsutil cp $path $workingDir/$metadataFileName
+
+    # iterate over config for dataset
+    props=$(jq -r ".\"$datasetId\" | keys[]"  <<< "$layersConfig")
+    for prop in ${props}
```
pwambach, line 36: seems to be possible to merge 2 objects with jq https://stackoverflow.com/a/24904276
but i'm also fine with this solution

scripts/update-metadata.sh
pmast: Yes, we could do this. But I am caching the content to save i/o operations probably overcautious?!

scripts/update-metadata.sh
pmast: Well found!

I will give the merge a try tomorrow, would save us the nested loop and fiddling around with tmp files.

scripts/update-metadata.sh
pmast: Using the file directly for readability

scripts/update-metadata.sh
pmast: Done!

PR #436, opened 2020-07-27
fix(sharing): fix sharing links and icon

### LINKED ISSUE #419: Sharing: Fix twitter icon
The bird is missing its beak:

image

### LINKED ISSUE #425: Sharing: implement
- [ ] Facebook: see here: https://github.com/ubilabs/mdr-klimakarte/blob/023f559cfbe146f8fd7d78954bd55f1b1fa6dfc7/src/components/ShareButton.js#L38-L46
  - this is officially not supported anymore, but unofficially this works since years, should be fine for now
- [ ] Twitter: see here: https://github.com/ubilabs/mdr-klimakarte/blob/023f559cfbe146f8fd7d78954bd55f1b1fa6dfc7/src/components/ShareButton.js#L28-L36
- [ ] add open graph fields for proper sharing: https://github.com/ubilabs/mdr-klimakarte/blob/023f559cfbe146f8fd7d78954bd55f1b1fa6dfc7/src/templates/index.html#L9-L14
  - use english texts
  - we will add an image later
- [ ] implement copying

Testing open graph fields: https://developers.facebook.com/tools/debug/

KatvonRivia:
closes #419 
closes #425


src/scripts/components/icons/twitter-icon.tsx
```diff
@@ -2,10 +2,14 @@ import React, {FunctionComponent} from 'react';
 
 export const TwitterIcon: FunctionComponent = () => (
   <svg
-    width="24"
-    height="24"
-    viewBox="0 0 24 20"
+    width="27"
```
pwambach, line 5: Could you please check the icons again, something seems a bit off here. I guess it's because the copy icon is not vertically centered. Also a bit more space between the buttons would look better imo.
Screenshot 2020-07-29 at 08 35 35

src/index.html
```diff
@@ -4,6 +4,21 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <!-- OpenGraph -->
+    <meta property="og:title" content="ESA - Climate from space" />
+    <meta property="og:type" content="website" />
+    <meta
+      property="og:image"
+      content="https://storage.googleapis.com/esa-cfs-storage/0.6.1/stories/a-passage-opens/assets/sst_large_18.jpg"
```
pwambach, line 12: We need to make these urls dynamic. As this has to be in place before any javascript kicks in we have to insert the correct url at build time.
One solution would be to set the environment variable NODE_ENV to 'live' and then insert the "live" url. We can discuss this in person if you like...

PR #448, opened 2020-08-06
feat(welcomeScreen): add custom story markers to globe

### LINKED ISSUE #95: welcome screen
As a welcome screen show markers with story names to trigger the user to dive into the stories right away
- [ ] show a pin for each stories
- [ ] show the name of the story
- [ ] clicking the pin enters the story
- [ ] let the globe spin slightly

KatvonRivia:
closes #95

Kapture 2020-08-10 at 16 24 34


src/scripts/libs/create-markers.ts
```diff
@@ -0,0 +1,214 @@
...
+  const blob = await response.blob();
+  const reader = new FileReader();
+
+  fontPromise = new Promise(resolve => {
+    reader.onload = () => {
+      resolve(reader.result as string);
+    };
+    reader.readAsDataURL(blob);
+  });
+  return fontPromise;
```
pwambach, line 64: new empty line please

src/scripts/components/globe/globe.tsx
```diff
@@ -295,6 +299,8 @@ const Globe: FunctionComponent<Props> = ({
     flyToGlobeView(viewer, flyTo);
   }, [viewer, flyTo]);
 
+  useMarkers(viewer, stories);
```
pwambach, line 302: If the logic is that we only show the markers when there is no story or data layer, I think a clearer solution would be to give the globe a property 'markers' and pass some marker data when there is no story pr data layer active. Sorry that my first guess was to use a flag in the redux store but with this behaviour it doesn't seem like a robust solution anymore.

We can talk about this in person if you want to...

src/scripts/components/app/app.styl
```diff
@@ -5,6 +5,11 @@
   font-family: NotesEsa
   src: url('../../../../assets/fonts/NotesEsaReg.otf')
 
+@font-face
```
pwambach, line 8: do we still need this here? we only use the bold font in the markers which load the font by themselves, right?

src/scripts/components/app/app.tsx
```diff
@@ -36,6 +38,19 @@ const App: FunctionComponent = () => (
 
 const TranslatedApp: FunctionComponent = () => {
   const language = useSelector(languageSelector);
+  const selectedLayers = useSelector(selectedLayerIdsSelector);
```
pwambach, line 41: we could put this stuff in a custom hook 'const markers = useStoryMarkers()' so that we keep the `app.tsx` clean

pwambach: also only map over the stories if hideMarkers is false to save some work if we don't show the markers

src/scripts/components/app/app.styl
KatvonRivia: Isn't this the global font import for all components? We use both fonts in other components.

KatvonRivia: Sorry i mean: you're right we don't use is yet 😄 But we need it at least for the splash screen title. I have used the regular one before, but that's not correct.

pwambach: ah I see 👍

src/scripts/hooks/use-story-markers.ts
```diff
@@ -0,0 +1,24 @@
...
+import {StoriesStateSelector} from '../selectors/story/story-state';
+
+export const useStoryMarkers = () => {
+  const selectedLayers = useSelector(selectedLayerIdsSelector);
+  const stories = useSelector(StoriesStateSelector).list;
+  const hideMarkers = Boolean(
+    selectedLayers.mainId || selectedLayers.compareId
+  );
+
+  if (!hideMarkers) {
```
pwambach, line 13: could you turn this around. we usually always use the positiv case:
```
if (hideMarkers) {
  return []
}

return ... storyMarkers
```

PR #461, opened 2020-08-12
feat(time-slider): improve time slider

### LINKED ISSUE #428: Timeslider: fixes/tweaks
- [x] in single data view, center the date above the handle
- [x] in compare mode
  - move the top date (for the left globe) to the left of the handle and align right with the handle
  - move the bottom date (for the right globe) to the right of the handle and align left with the handle
- [x] move the min/max dates between the two rails in compare mode
- [x] adjust colors of the track and the data points to closer match the designs in figma:

image

pwambach:
closes #428

### DIFF
--- a/src/scripts/components/time-slider-range/time-slider-range.tsx
+++ b/src/scripts/components/time-slider-range/time-slider-range.tsx
@@ -29,8 +29,8 @@ const TimeSliderRange: FunctionComponent<Props> = ({
 
     return {
       left: `${tickPosition}%`,
-      backgroundColor: isSelected ? 'white' : undefined, // eslint-disable-line no-undefined
-      transform: isSelected ? 'translateX(-50%) scale(2)' : undefined // eslint-disable-line no-undefined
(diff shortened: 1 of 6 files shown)


src/scripts/components/time-slider-range/time-slider-range.tsx
```diff
@@ -29,8 +29,8 @@ const TimeSliderRange: FunctionComponent<Props> = ({
 
     return {
       left: `${tickPosition}%`,
-      backgroundColor: isSelected ? 'white' : undefined, // eslint-disable-line no-undefined
-      transform: isSelected ? 'translateX(-50%) scale(2)' : undefined // eslint-disable-line no-undefined
+      backgroundColor: isSelected ? '#00AE9D' : undefined, // eslint-disable-line no-undefined
```
KatvonRivia, line 32: Is it possible to use the color variables here as well?

pwambach: unfortunately not, the stylus variables are not accessible from javascript :(

PR #483, opened 2020-08-14
refactor(story): add layer description, hide header title on splashsc…

### LINKED ISSUE #432: Story mode: Allow globe views to have subtitles
- See figma for designs
- Take care of the necessary data structure
- Update the template spreadsheet with an example.

### LINKED ISSUE #451: Stories: slide add padding bottom to show all content
Currently the bottom part of the slides is hidden behind the footer:

image

- [ ] show all content of slide

### LINKED ISSUE #459: Splash screen: remove story title from header
Remove the story title from the header (for the splash screen only).
The title will be present on the splash screen.

b1 Story Page 1

### LINKED ISSUE #463: Stories: change body background to black
- [ ] change the body background of the story panels to black
- [ ] change the background of the story selector to black

This helps images and video blend in more nicely.

KatvonRivia:
…reen

closes #432 
closes #463 
closes #459 
closes #451


pwambach:
Another issue here is that we cannot just change the story jsons because now the content comes from the partners. All changes to properties have to be reflected in the story template sheet https://docs.google.com/spreadsheets/d/1kL8UYm5oFuKnN97hvPVp9Oj6YQ2q6CZxo_wCr9Qs5sQ/edit#gid=0 
In addition the story mapper has to be adjusted when adding new fields.

KatvonRivia:
> Could you please split this up in multiple PRs. 22 file changes which close 4 tickets is a bit confusing to review. Also what are the changes on the image files and what are the additional .gstmp files you added?
Hmm i didn't add these files, but i downloaded everything from storage before i added the layerDescriptions. Maybe it came from there. 🤔

src/scripts/components/main/app/app.tsx
```diff
@@ -60,7 +60,7 @@ const TranslatedApp: FunctionComponent = () => {
             <div className={styles.logo}>
               <EsaLogo />
             </div>
-            <Globes markers={markers} />
+            <Globes markers={markers} backgroundColorString={'#10161A'} />
```
pwambach, line 63: sorry for another comment 🙈  but could you remove the "String" from "backgroundColorString". the typescript type already defines that it is a string

KatvonRivia: Haha no problem. I named it _backgroundColorString_ because i'm using _backgroundColor_ already in globe.tsx:
`const backgroundColor = Color.fromCssColorString(backgroundColorString);`

How about something like backgroundColorCode or backgroundColorHex, then?

pwambach: I would prefer to rename the inner variable instead. The property names are more important to be precise because it's the interface for someone who uses the component and doesn't (need to) know whats going on inside. 
Also in this case the inner variable is not needed to be defined as it is only used once. Just set the color directly:
```js
scopedViewer.scene.backgroundColor = Color.fromCssColorString(backgroundColorString);
```

KatvonRivia: True, good point. :) I'll change it, thanks.

pwambach (approved):

PR #487, opened 2020-08-19
feat(mobile-stories): add mobile compatibility

KatvonRivia:


src/scripts/components/stories/story/story.styl
```diff
@@ -13,3 +13,19 @@
...
+.rightSideComponent
+  height: 100%
+
+@media screen and (max-width: 480px)
+  .main
+    display: flex
+    flex-direction: column
+    overflow-y: auto
+
+    > :last-child
```
pwambach, line 26: do we have a className for this element?

KatvonRivia: Sadly no, it can be either the gallery, story video or the globe.

pwambach: ah ok, then 👍

PR #490, opened 2020-08-19
feat(mobile-data-mode): add mobile compatibility

KatvonRivia:


src/scripts/components/main/app/app.styl
```diff
@@ -33,7 +33,12 @@
...
-  top: emCalc(22px)
+  top: emCalc(19px)
   left: emCalc(22px)
   z-index: 4
+  pointer-events: none
   fill: $textWhite
+
+@media screen and (max-width: 480px)
+  :global(path.hideText)
+    display: none !important
```
pwambach, line 44: I think we can remove the important here (?)

src/scripts/components/main/button/button.styl
```diff
@@ -36,3 +36,6 @@
   .hideLabel
     span
       display: none
+
+  button.button.hideLabel
```
pwambach, line 40: why the `button.button.`? Isn't this the same class as the hideLabel above?

pwambach (changes requested):

src/scripts/components/main/button/button.styl
KatvonRivia: It gets overwritten by 

`  .button, &:not(:first-child)
    padding-left: emCalc(30px)` 

in navigation.tsx, so i needed to give it a higher specification.

pwambach: Looks to me like the problem should be solved in the navigation component. Not sure if I understand the problem but the "shared" button component shouldn't have to adjust its code because someone who uses the component overwrites its styles.

KatvonRivia: Makes sense, thanks! I will change it.

pwambach (approved):

PR #506, opened 2020-08-24
refactor(filter): browse through filters by clicking

KatvonRivia:


src/scripts/components/stories/story-filter/story-filter.tsx
```diff
@@ -44,52 +42,28 @@ const StoryFilter: FunctionComponent = () => {
...
-  }, [scrollLeft]);
-
-  // handle right side scrolling
-  useEffect(() => {
-    if (!scrollRight) {
-      return () => {};
+  const leftClick = () => {
+    if (innerRef.current && translateValue < 0) {
+      translateValue += scrollSpeed;
+      innerRef.current.style.transform = `translateX(${translateValue +
```
pwambach, line 48: why do you add the scrollSpeed here again?

src/scripts/components/stories/story-filter/story-filter.tsx
```diff
@@ -13,18 +13,16 @@ import setSelectedStoryTags from '../../../actions/set-selected-story-tags';
...
   const dispatch = useDispatch();
   const stories = useSelector(storyListSelector);
   const selectedTags = useSelector(selectedTagsSelector);
   const innerRef = useRef<HTMLDivElement>(null);
-  const [scrollLeft, setScrollLeft] = useState(false);
-  const [scrollRight, setScrollRight] = useState(false);
-  const maxScroll =
-    (innerRef.current?.scrollWidth || 0) - (innerRef.current?.offsetWidth || 0);
-
+  const getMaxScroll = () =>
```
pwambach, line 23: why is this a function now? shouldn't this value be static?

src/scripts/components/stories/story-filter/story-filter.tsx
```diff
@@ -44,52 +42,28 @@ const StoryFilter: FunctionComponent = () => {
...
-      if (innerRef.current && translateValue > maxScroll * -1) {
+  const rightClick = () => {
+    if (innerRef.current) {
+      const maxScroll = innerRef.current && getMaxScroll();
+      if (translateValue >= maxScroll * -1) {
         translateValue -= scrollSpeed;
-        innerRef.current.style.transform = `translateX(${translateValue}px)`;
-        raf.id = requestAnimationFrame(loop);
+        innerRef.current.style.transform = `translateX(${translateValue +
+          scrollSpeed}px)`;
```
pwambach, line 59: same here?

src/scripts/components/stories/story-filter/story-filter.tsx
```diff
@@ -13,18 +13,16 @@ import setSelectedStoryTags from '../../../actions/set-selected-story-tags';
 import styles from './story-filter.styl';
 
 let translateValue = 0;
```
pwambach, line 15: when we don't use this in a requestAnimationFrame loop, this value should be in a useState variable and use the normal react render mechanism

pwambach (changes requested):
When changing this to react on clicks the whole component should have another behaviour. The

pwambach (approved):

PR #515, opened 2020-08-24
refactor(story-markers): use markers on story globe

### LINKED ISSUE #465: Story: allow to add pins
Pins in the context of stories will be used to mark/highlight certain areas and over additional data for download.

- [ ] Allow the globe in the stories to have pins.
- [ ] pins can have labels
- [ ] use the same pins as we have for the welcome screen
- only for single globe mode

KatvonRivia:
closes #465


src/scripts/hooks/use-markers.ts
```diff
@@ -27,12 +31,13 @@ export const useMarkers = (viewer: Viewer | null, markers: Marker[]) => {
       scene.canvas as HTMLCanvasElement
     );
 
-    handler.setInputAction(movement => {
-      const pickedObject = scene.pick(movement.position);
-      if (defined(pickedObject)) {
-        history.push(`/stories/${pickedObject.id._id}/0`);
-      }
-    }, ScreenSpaceEventType.LEFT_CLICK);
+    markerLink &&
```
pwambach, line 34: Thinking about it the `markerLink` property is not really a nice way to solve this. It would be cleaner and more flexible if we add a `link` property to each marker. You can add properties to cesium entities with https://cesium.com/docs/cesiumjs-ref-doc/Entity.html#properties
This way each marker can decide if it wants to be a link or not (by nor setting the link propery)

pwambach (changes requested):

pwambach (approved):

PR #518, opened 2020-08-25
refactor(marker): make id optional

KatvonRivia:


src/scripts/libs/create-marker.ts
```diff
@@ -31,7 +31,7 @@ export async function createMarker(marker: Marker): Promise<Entity> {
 
       resolve(
         new Entity({
-          id: `${marker.id}`,
+          id: marker.id && `${marker.id}`,
```
pwambach, line 34: does this work without any id? if not we could set the title to have something uniq as the id

pwambach (approved):

src/scripts/libs/create-marker.ts
KatvonRivia: Yes it creates a unique id automatically if none is set.

PR #522, opened 2020-08-25
Feat/electron cosmetics

### LINKED ISSUE #454: Standalone app
- [x] give the app an icon to show up nicely in the app switcher menu
- [x] gif the App an name to show up nicely in the app switcher: "ESA – Climate from Space"
- [x] do not open dev tools on startup

mrMetalWood:
- icon for app switcher
- nicer name for app and window
- don't open devtools in production
- handle deprecation warning

esa

closes #454


src/electron/main.ts
```diff
@@ -12,10 +13,12 @@ function createWindow() {
   const window = new BrowserWindow({
     width: 1400,
     height: 800,
+    title: 'ESA – Climate from Space',
     webPreferences: {
       nodeIntegration: false,
       contextIsolation: true,
-      preload: path.join(__dirname, 'preload.js')
+      preload: path.join(__dirname, 'preload.js'),
+      enableRemoteModule: true
```
pwambach, line 21: why do you enable the remote module? it is deprecated and should not be used. see https://www.electronjs.org/docs/breaking-changes#default-changed-enableremotemodule-defaults-to-false

pwambach (changes requested):

src/electron/main.ts
mrMetalWood: Oh, I thought we need it and wanted to get rid of the warning. Why else would it warn me, if I didn't use it? Will remove it.

pwambach: Yeah I was confused here too at first, but they added the warning because they changed the default value for the property which could break a lot of apps.

mrMetalWood: Keeping it, since electron 10 default is set to false and its not trivial to move away from it.

pwambach: Added a new ticket #527  to remove the remote module

pwambach (approved):

PR #526, opened 2020-08-25
feat(check-language): add info bubble and store language in localStorage

### LINKED ISSUE #466: Chore: on initial load check langauge
- [ ] on initial load ask whether user is fine with language selection
- [ ] keep selected language for future visits
- [ ] add a language icon in front of the text

image

KatvonRivia:
closes #466


src/scripts/components/main/app/app.tsx
```diff
@@ -36,12 +37,21 @@ const App: FunctionComponent = () => (
 );
 
 const TranslatedApp: FunctionComponent = () => {
-  const language = useSelector(languageSelector);
   const markers = useStoryMarkers();
+  const language = useSelector(languageSelector);
+
+  const getLanguage = () => {
+    if (localStorage.getItem('language')) {
+      return localStorage.getItem('language') as Language;
```
pwambach, line 45: We already retrieve the browser language in the language reducer. This would be a good place to also check for the localStorage language. It also has the benefit that it is only evaluated once on app start and not on every render like here. 

Also we have to check if the language in the storage is actually one of the available languages similar to the getBrowserLanguage function.

src/scripts/components/main/language-selector/language-selector.tsx
```diff
@@ -21,6 +21,14 @@ const LanguageSelector: FunctionComponent<Props> = ({className = ''}) => {
   const selectedLanguage = useSelector(languageSelector);
   const dispatch = useDispatch();
 
+  const setLanguage = (language: Language) => {
+    localStorage.setItem('language', language);
+
+    if (localStorage.getItem('language')) {
```
pwambach, line 27: what is this check for? if you set the item first and then check if it's there it should be always true

src/scripts/components/main/navigation/navigation.tsx
```diff
@@ -11,10 +11,20 @@ import Share from '../share/share';
 import {MenuIcon} from '../icons/menu-icon';
 
 import styles from './navigation.styl';
+import InfoBubble from '../info-bubble/info-bubble';
+import {languageSelector} from '../../../selectors/language';
 
 const Navigation: FunctionComponent = () => {
   const dispatch = useDispatch();
   const [showMenu, setShowMenu] = useState(false);
+  const [showInfoBubble, setshowInfoBubble] = useState(true);
```
pwambach, line 20: We don't want to show the bubble if there is already a language in local storage right?
Something like that?
```js
const savedLanguage = localStorage.getItem('language');
const [showInfoBubble, setshowInfoBubble] = useState(!Boolean(savedLanguage));
```

src/scripts/components/main/language-selector/language-selector.tsx
```diff
@@ -21,6 +21,14 @@ const LanguageSelector: FunctionComponent<Props> = ({className = ''}) => {
   const selectedLanguage = useSelector(languageSelector);
   const dispatch = useDispatch();
 
+  const setLanguage = (language: Language) => {
+    localStorage.setItem('language', language);
```
pwambach, line 25: It would be cool if the saving to localStorage and the dispatch would be combined so that we don't have to handle the saving manually in multiple places. We could just always save the language to local storage when dispatching the `setLanguageAction` action in the `actions/set-language.ts` file

pwambach (changes requested):

src/scripts/config/main.ts
```diff
@@ -78,5 +78,6 @@ export default {
       'http://twitter.com/intent/tweet?text=ESA%20Climate%20From%20Space&url={currentUrl}'
   },
   legendImage: `${baseUrlStorage}/legend-images/{variable}.png`,
-  downloadUrls
+  downloadUrls,
+  localStorageLanguage: 'language'
```
pwambach, line 82: It's the key and not the language itself. Please name it 'localStorageLanguageKey'

src/scripts/components/main/app/app.tsx
```diff
@@ -39,19 +38,11 @@ const App: FunctionComponent = () => (
...
   const markers = useStoryMarkers();
   const language = useSelector(languageSelector);
-
-  const getLanguage = () => {
-    if (localStorage.getItem('language')) {
-      return localStorage.getItem('language') as Language;
-    }
-    return language;
-  };
+  localStorage.clear();
```
pwambach, line 41: Why did you add this? Debugging I guess, please remove.

src/scripts/libs/get-local-storage-language.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export default function getLocalStorageLanguage(): Language | null {
+  const storedLanguage = localStorage.getItem(config.localStorageLanguage);
+  const languages = Object.values(Language);
+
+  const availableLanguage =
+    storedLanguage &&
+    languages.find(
+      language =>
+        language.substr(0, 2).toLowerCase() ===
```
pwambach, line 13: We don't need the substr stuff here. Unlike in the getBrowserLanguage function the language in the localStorage should match the defined languages exactly.

pwambach (changes requested):

pwambach (approved):

PR #533, opened 2020-08-26
feat(globe): add auto rotate on initial load

### LINKED ISSUE #14: slightly spinning globe on landing page
- Slightly spin the globe  when opening the application
- In the correct direction!

mrMetalWood:
- slightly spin the globe on initial load and when switching from 2D to 3D again
- stop spinning on first mouseDown event on the globe
- stop spinning when switching to 2D and when clicking on the flyTo compass thingy
- add spinning flag to globe url state and also restore from url

spinning

closes #14


src/scripts/components/main/globe/globe.tsx
```diff
@@ -68,19 +79,23 @@ function getBasemapUrl(id: BasemapId | null) {
   return isElectron() ? config.basemapUrlsOffline[id] : config.basemapUrls[id];
 }
 
+const spinningEventHelper = new EventHelper();
```
pwambach, line 82: Why do we need this? Can't we just attach the handler directly to the clock?
And if we need it, shouldn't it be scoped inside the component so that multiple globe instances don't share the same object

pwambach (changes requested):
Very nice, just one small issue...

pwambach (approved):

PR #541, opened 2020-08-27
feat(layer-info): content updates

pmast:
Might be controversial: added HTML support for React markdown in the layers-info to support the extended legend for landcover:

image


mrMetalWood:
Not sure about the HTML support and its implications. Lets see what @pwambach says :) Fine with me

mrMetalWood (approved):

pwambach:
Actually I would have preferred to not enable html injects because it's always a security issue to insert html from a remote server. In our case it's not user generated so it "should" be ok, but you never know...

pmast:
I can also try and work with an image instead. I will give it a try later.

pwambach:
But we would need one image per language :(

pmast:
True…

pmast:
Is there an option for screenshotting HTML elements with transparent background?

PR #549, opened 2020-08-28
refactor(layer-list): shorten layer names on hover

### LINKED ISSUE #498: When switching data layers…
- [ ] Stop the playback when switching the data layers.

### LINKED ISSUE #548: Layer Selector: shorten names on hover
- [ ] shorten Layer names with an ellipsis on hover when they would break to the second line because of the 'compare' text.

image

KatvonRivia:
closes #548

### DIFF
--- a/src/scripts/components/layers/layer-list-item/layer-list-item.styl
+++ b/src/scripts/components/layers/layer-list-item/layer-list-item.styl
@@ -10,6 +10,13 @@
   &:hover, &:focus
     background-color: rgba($darkGrey4, 0.5)
 
+    .layerTitle
+      display: block
+      overflow: hidden
+      width: 100%
+      text-overflow: ellipsis
+      white-space: nowrap
+
     .compare
       display: block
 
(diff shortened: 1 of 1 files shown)


src/scripts/components/layers/layer-list-item/layer-list-item.styl
```diff
@@ -10,6 +10,13 @@
   &:hover, &:focus
     background-color: rgba($darkGrey4, 0.5)
 
+    .layerTitle
```
mrMetalWood, line 13: I think with CSS Modules you could to something like this to avoid the repetition:

Somewhere at the top:
```css
.ellipsis
  display: block
  overflow: hidden
  width: 100%
  text-overflow: ellipsis
  white-space: nowrap
```

And then
```css
.layerTitle
  composes: ellipsis
```

mrMetalWood (approved):

src/scripts/components/layers/layer-list-item/layer-list-item.styl
KatvonRivia: Ah truee, wie used something similar in the slider.

PR #550, opened 2020-08-28
Feat/gallery layer type

### LINKED ISSUE #535: Add profile base data sets
Add an explorer for profile data.

Example:

image

- Data: https://drive.google.com/drive/folders/1xT8cJaO13vFTlqOkJqWuLlvdHKBWckL7
- take daily resolution from this data
- open images instead of globe
- link data display to time slider

- keep legend
- keep time slider
- do not show globe
- do not show globe controls

Example Meta Data file:
https://storage.googleapis.com/esa-cfs-tiles/0.7.1/aerosol.AOD550_mean/metadata.json

image

mrMetalWood:
- adds new gallery layer type
- renamed `globes.tsx` to  `data-viewer.tsx` to be able to show globes as well as the new gallery layer type
- adds `ozone_profile` layer which makes use of the new layer type

single
Bildschirmfoto 2020-08-28 um 16 23 53

compare
Bildschirmfoto 2020-08-28 um 16 25 30

close #535


src/scripts/components/main/data-viewer/data-viewer.tsx
```diff
@@ -23,16 +24,23 @@ import {globeSpinningSelector} from '../../../selectors/globe/spinning';
 
 import {GlobeView} from '../../../types/globe-view';
 import {Marker} from '../../../types/marker-type';
+import {LayerType} from '../../../types/globe-layer-type';
 
-import styles from './globes.styl';
+import styles from './data-viewer.styl';
```
pwambach, line 29: Can you please move the styles import to the end. We tried to have the same import structure everywhere. Which is:
* node modules/native imports
* own modules
* types
* style imports

each block separated with a new line

src/scripts/components/main/data-viewer/data-viewer.tsx
```diff
@@ -87,59 +95,79 @@ const Globes: FunctionComponent<Props> = ({backgroundColor, markers = []}) => {
...
-      )}
-      {compareLayerDetails && (
-        <LayerLegend
-          id={compareLayerDetails.id}
-          values={[compareLayerDetails.maxValue, compareLayerDetails.minValue]}
-          unit={compareLayerDetails.units}
-          isCompare={true}
-        />
-      )}
+  const showGlobeNavigation =
```
pwambach, line 98: Please add a comment for this check

src/scripts/components/main/data-viewer/data-viewer.tsx
```diff
@@ -87,59 +95,79 @@ const Globes: FunctionComponent<Props> = ({backgroundColor, markers = []}) => {
...
         onMoveEnd={onMoveEndHandler}
         onMouseDown={onMouseDownHandler}
       />
+    );
+  };
+
+  return (
+    <div className={styles.dataViewer}>
+      {[mainLayerDetails, compareLayerDetails]
+        .filter((layer): layer is Layer => Boolean(layer))
```
pwambach, line 142: is this the same as `.filter((layer: Layer) => Boolean(layer))`?

src/scripts/components/main/data-viewer/data-viewer.tsx
```diff
@@ -87,59 +95,79 @@ const Globes: FunctionComponent<Props> = ({backgroundColor, markers = []}) => {
...
         projectionState={projectionState}
-        imageLayer={mainImageLayer}
-        basemap={mainLayerDetails?.basemap || null}
+        imageLayer={imageLayer}
+        basemap={layerDetails?.basemap || null}
         spinning={globeSpinning}
         flyTo={flyTo}
-        onMouseEnter={() => setIsMainActive(true)}
-        onTouchStart={() => setIsMainActive(true)}
+        onMouseEnter={action}
```
pwambach, line 130: I guess this action can be hardcoded because it has to be this one for the globe and it saves one parameter :)

pwambach (changes requested):
Cool! Just some minor stuff...

src/scripts/components/main/data-viewer/data-viewer.tsx
mrMetalWood: No it's not. I was fighting typescript on this one. I never understand why typescript (and also flow) can not infer this kind of stuff. I clearly filter out all the `null` layers and then it complains in the VERY NEXT LINE that it could be null :D

Bildschirmfoto 2020-08-31 um 11 56 28

src/scripts/components/main/data-viewer/data-viewer.tsx
mrMetalWood: Where would you hardcode it? Its `() => setIsMainActive(true)` for the main globe and `() => setIsMainActive(false)` for the compare globe.

src/scripts/components/main/data-viewer/data-viewer.tsx
mrMetalWood: I didn't know this pattern before. Its called type predicates or custom type guards. https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

pwambach: ah ok, yes strange that this does not work out of the box

src/scripts/components/main/data-viewer/data-viewer.tsx
pwambach: I wouldn't make it a parameter of the `getDataWidget` function and write in line 130 `onMouseEnter={() => setIsMainActive(true)}`

mrMetalWood: But then it would be the same action for both globes. At the moment one sets it `true` and the other `false`

pwambach: Ah i got this wrong :) Makes sense!

pwambach (approved):

PR #576, opened 2020-09-02
Feat/timeslider tweaks

### LINKED ISSUE #498: When switching data layers…

### LINKED ISSUE #510: Time slider: issues
- [x] when changing datasets, time slider seems to be stuck in different times
- [x] set new playback speed: constant per tick, when two datasets are show, use smaller ticks as reference
- [x] check time slider on Firefox

mrMetalWood:
- clamp time to range of active layers when a layer changes
- use individual playback steps for each dataset, use smaller one when two are active
- fix firefox styling of timeslider
- stop playback when layers change

closes #510 closes #498


src/scripts/components/layers/time-slider/time-slider.tsx
```diff
@@ -46,6 +49,11 @@ const TimeSlider: FunctionComponent = () => {
     layerDetailsSelector(state, compareId)
   );
 
+  const playbackStep = useMemo(
```
pwambach, line 52: Cool

src/scripts/components/layers/time-slider/time-slider.tsx
```diff
@@ -79,6 +87,12 @@ const TimeSlider: FunctionComponent = () => {
     []
   );
 
+  // clamp globe time to min/max of the active layers when a layer changes and stop playback
+  useEffect(() => {
```
pwambach, line 91: Don't we have the combined time ranges with a min/max value already in the timeslider? Why the extra action and not just call `setGlobeTime(clampToRange(time, combined.min, combined.max))` directly?

mrMetalWood: yep, correct. Will change that.

pwambach (approved):

PR #582, opened 2020-09-03
refactor(globe): only rerender when data actually changes

### LINKED ISSUE #557: Time Slider: gaps in ticks are not shown
The sea ice layer has a gap from October 2011 to August 2012.
- [x] The time slider track should show a gap where there is no data.
- [x] also the data set seems to change when the handle is moved through the gap

mrMetalWood:
Prevent rerenders when the data didn't actually change. Right now the layers are re-rendering on every time change since the variable reference changes.
The imageLayer is a fairly small object and can safely be stringified and parsed I would say.

closes #557


pwambach (changes requested):
Nice finding! But the JSON.stringify feels like a hack. I think the clean way to solve this would be to put the 
```
const mainImageLayer = useMemo(
    () => getImageLayerData(mainLayerDetails, time),
    [mainLayerDetails, time]
  );
```

logic from the data-viewer in a custom hook where the useMemo doesn't depend on `time` but on the 

```
const timeIndex = getLayerTimeIndex(time, layer.timestamps);
```
variable. This one only changes when there is new data.

pwambach (approved):

PR #643, opened 2020-09-18
feat(attributions): add attributions to menu

KatvonRivia:


src/scripts/components/main/menu/menu.tsx
```diff
@@ -15,11 +15,13 @@ import {LinuxIcon} from '../icons/linux-icon';
 import {AppleIcon} from '../icons/apple-icon';
 import config from '../../../config/main';
 import {Ubilabslogo} from '../icons/ubilabs-logo';
+import Attributions from '../attributions/attributions';
 
 import styles from './menu.styl';
 
 const Menu: FunctionComponent = () => {
   const [showOverlay, setShowOverlay] = useState(false);
+  const [overlayContentType, setOverlayContentType] = useState('');
```
pwambach, line 24: One variable should be enough here like 
```
const [overlayType, setOverlayType] = ...
setOverlayType('about')
setOverlayType('attributions')
setOverlayType(null) // to hide the overlay
```

pwambach (changes requested):

src/scripts/components/main/menu/menu.tsx
KatvonRivia: oh true 🙈  thanks!

src/scripts/components/main/menu/menu.tsx
```diff
@@ -20,21 +20,16 @@ import Attributions from '../attributions/attributions';
 import styles from './menu.styl';
 
 const Menu: FunctionComponent = () => {
-  const [showOverlay, setShowOverlay] = useState(false);
-  const [overlayContentType, setOverlayContentType] = useState('');
+  const [overlayType, setOverlayType] = useState<string | null>('');
```
pwambach, line 23: the default value should be `null` not `''`

pwambach (changes requested):

pwambach (approved):

PR #652, opened 2020-09-22
feat(stories): add data download

pmast:
Adding the possibility to link with relative links to files in the story's assets folder.
Had to add the `transformLinkUri` function to reactmarkdown.

image


src/scripts/libs/get-story-urls.ts
```diff
@@ -0,0 +1,40 @@
...
+  getOfflineStoryMediaUrl
+} from './electron/index';
+import {replaceUrlPlaceholders} from './replace-url-placeholders';
+import config from '../config/main';
+
+export function getStoryMediaUrl(
+  storyId: string,
+  relativePath: string
+): string {
+  return getAbsolutePath(storyId, relativePath);
```
pmast, line 13: There is probably a more elegant way to solve this. Let me know please.

pwambach:
What was the problem here? Isn't it possible to just use relative paths in the markdown link like e.g. `./assets/csvs/file1.csv` ?

pmast:
No that does not work as I would keep those files in a story specific assets folder.

The relative URL resolve to

`localhost:8080/assets/ee-chart.csv`

But should resolve to:

`localhost:8080/storage/story-16/assets/ee-chart.csv`

pwambach:
Ah now I get it. Looks fine to me. The only thing I wonder is do we have to separate the two functions for images and links? Seems like the 'http' check would make sense for both.

pwambach:
Also some eslint errors involved :)
```
/workspace/src/scripts/libs/get-story-urls.ts
  34:1  error  Trailing spaces not allowed  no-trailing-spaces
  38:1  error  Trailing spaces not allowed  no-trailing-spaces
```

pwambach (approved):

PR #655, opened 2020-09-23
refactor(ubilabs-logo): change size and color

### LINKED ISSUE #646: Ubilabs attribution: update
- Make logo white
- Link logo to website
- left align logo and text
- and make it larger

(and/or briefly ask Elena for a sketch)

KatvonRivia:
closes #646

### DIFF
--- a/src/scripts/components/main/menu/menu.styl
+++ b/src/scripts/components/main/menu/menu.styl
@@ -60,7 +60,6 @@
   display: flex
   flex-direction: row
   justify-content: center
-  width: 100%
 
   svg
     fill: $main
@@ -77,10 +76,9 @@
   bottom: emCalc(20px)
   left: emCalc(32px)
   color: $textDefault
-  font-size: 10px
 
   svg
-    height: remCalc(25px)
+    height: emCalc(25px)
     fill: $textWhite
 
     &:hover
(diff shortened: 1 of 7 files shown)


src/scripts/components/main/menu/menu.styl
```diff
@@ -80,15 +80,16 @@
   font-size: 10px
 
   svg
-    height: emCalc(30px)
-    fill: $textDefault
+    height: remCalc(25px)
```
pwambach, line 83: why rem? if the parent has `font-size: 10px` you can also use `emCalc(25px, 10)`

pwambach (changes requested):

src/scripts/components/main/menu/menu.styl
KatvonRivia: Why do you prefer that? I used rem because the size of the logo is not related to the font-size. It's not like a padding around text that should change with the font size. Maybe i should add a class to the version number, because the `font-size: 10px` is used only there. What do you think? 🤔

pwambach: usually you want all the child elements to scale with the parent if you increase the parent font-size. why the version number? the parent of the logo is the `.credits` element?

KatvonRivia: Ah, thats the problem 😄  i removed the font size from `.credits` it's not needed, sorry.

pwambach (approved):

PR #656, opened 2020-09-23
feat(autoplay-gallery): add autoplay to gallery in showcase mode

### LINKED ISSUE #638: Presenter mode: gallery
- [ ] play through gallery in show case mode
- [ ] in presenter mode clicks should progress through gallery before moving on to next slide.

KatvonRivia:
closes #638

### DIFF
--- a/src/scripts/components/stories/story-media/story-media.tsx
+++ b/src/scripts/components/stories/story-media/story-media.tsx
@@ -7,6 +7,7 @@ import {FullscreenExitIcon} from '../../main/icons/fullscreen-exit-icon';
 import {FullscreenIcon} from '../../main/icons/fullscreen-icon';
 import {getStoryMediaUrl} from '../../../libs/get-story-media-url';
 import {useInterval} from '../../../hooks/use-interval';
+import config from '../../../config/main';
 
 import {StoryMode} from '../../../types/story-mode';
 
@@ -32,6 +33,8 @@ const StoryMedia: FunctionComponent<Props> = ({
   const showPrevButton = currentIndex > 0;
   const showNextButton = currentIndex < images.length - 1;
 
+  const delay = mode === StoryMode.Showcase ? config.delay : null;
+
   useInterval(() => {
     if (mode === StoryMode.Showcase) {
       if (currentIndex >= images.length - 1) {
@@ -39,7 +42,7 @@ const StoryMedia: FunctionComponent<Props> = ({
       }
       setCurrentIndex(currentIndex + 1);
     }
-  }, 3000);
+  }, delay);
 
   const onPrevClick = () => {
     if (currentIndex <= 0) {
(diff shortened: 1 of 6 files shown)


src/scripts/hooks/use-story-navigation.ts
```diff
@@ -35,7 +38,12 @@ export const useStoryNavigation = () => {
   if (storyIds && typeof storyIndex === 'number') {
     const showcaseStoryIds = storyIds.join('&');
     const nextStoryIndex = storyIndex + 1;
+    const currentSlide = selectedStory?.slides[slideIndex];
     // go through all slides of one story
+
+    if (currentSlide?.images && currentSlide.type === SlideType.Image) {
+      delay = delay * currentSlide.images.length;
+    }
```
pwambach, line 46: add empty line

src/scripts/hooks/use-story-navigation.ts
```diff
@@ -15,6 +17,7 @@ export const useStoryNavigation = () => {
   let autoPlayLink = null;
   let nextSlideLink = null;
   let previousSlideLink = null;
+  let delay = 3000;
```
pwambach, line 20: Would be cool if this value is defined in the config file

src/scripts/components/stories/story-media/story-media.tsx
```diff
@@ -27,6 +32,15 @@ const StoryMedia: FunctionComponent<Props> = ({
   const showPrevButton = currentIndex > 0;
   const showNextButton = currentIndex < images.length - 1;
 
+  useInterval(() => {
```
pwambach, line 35: This way the interval function is called every 3 seconds also in non showcase stories. Better put the `mode === StoryMode.Showcase` check into the delay. If the delay is `null` the function will not be triggered.

Like:
```
delay = mode === StoryMode.Showcase ? 3000 : null
```

src/scripts/components/stories/story-media/story-media.tsx
```diff
@@ -6,19 +6,24 @@ import {NextIcon} from '../../main/icons/next-icon';
...
 import styles from './story-media.styl';
 
 interface Props {
   images: string[];
   imageCaptions?: string[];
   storyId: string;
+  mode: StoryMode | null;
 }
 
 const StoryMedia: FunctionComponent<Props> = ({
```
pwambach, line 22: Another thing but not related to this PR. It's a bit confusing that this component is called StoryMedia and not sth like StoryGallery or StoryImages. At first I thought StoryMedia is a wrapper component for images and videos etc.

pwambach (changes requested):

src/scripts/components/stories/story-media/story-media.tsx
KatvonRivia: Good point, will rename it in a different PR. 👍

pwambach (approved):

PR #691, opened 2020-10-15
fix(timeslider): show single slider when there is no main range

### LINKED ISSUE #651: Time slider: compare with biomass
- slider thinks there are two data layers, handle is not aligned correctly

KatvonRivia:
image

closes #651


mrMetalWood (approved):
nice :)

PR #718, opened 2020-10-21
refactor(story-selector): adjust spacing

KatvonRivia:


src/scripts/components/stories/story-list-item/story-list-item.styl
```diff
@@ -37,11 +37,8 @@
 
 .info
   overflow-y: auto
-  box-sizing: border-box
```
mrMetalWood, line 40: Is there a reason you removed this here? Often times unintuitive stuff happens when this is not set :D

mrMetalWood (approved):

src/scripts/components/stories/story-list-item/story-list-item.styl
KatvonRivia: Yeah you're right. The percentages also make more sense then.

PR #734, opened 2020-11-05
feat(tracking): add basic JS error tracking to matomo

### LINKED ISSUE #467: Error capturing
Capture errors that happen during runtime.
We have to report quarterly which errors happen.

Use User-Tracking platform to also store error events.

Catch global error events of app and send to tracking platform.

We do probably need user consent for sending this information.

mrMetalWood:
Also track JS errors to Matomo dashboard

closes #467


KatvonRivia (approved):
🚀

PR #807, opened 2020-12-15
feat(hover-legend): add legend with hover values

### LINKED ISSUE #620: Landcover: legend
Solution:
- [ ] on hover show corresponding value for each tick

---

Landcover has a very challenging legend:
- a lot of items
- long labels

Still we would like to show this at the regular place.
Currently wo do not show the label. We show only the colors.
We have the full long list in the data set description.

image
image

This is how a full list would look like:

image

Here also as HTML:

https://codepen.io/pmast/pen/XWdyzvg

Link to the data: https://storage.googleapis.com/esa-cfs-versions/web/develop/index.html#/?globe=SI9.62I46.57I18797866.86I360.00I-89.61I0.00I1420070400000.00I0.00Iland_cover.lccs_classI

KatvonRivia:
closes #620


pmast:
It would be good to give the hovered element some more highlighting (an outline). No the enlarging effect seems too arbitrary.

src/scripts/components/layers/layer-legend/layer-legend.tsx
```diff
@@ -1,4 +1,4 @@
-import React, {FunctionComponent} from 'react';
+import React, {FunctionComponent, useState} from 'react';
```
pwambach, line 1: I think it would be cleaner to use a separate component for this type of legend. Would be less confusing and who knows which other legend types will come in the future

src/scripts/components/layers/layer-loader/layer-loader.tsx
```diff
@@ -20,7 +20,6 @@ const LayerLoader: FunctionComponent = () => {
   const compareLayerDetails = useSelector((state: State) =>
     layerDetailsSelector(state, compareId)
   );
-
   // load layer list on mount
```
pwambach, line 23: Please leave the extra line in there

src/scripts/config/main.ts
```diff
@@ -89,5 +89,104 @@ export default {
   legendImage: `${baseUrlTiles}/{id}/legend.png`,
   downloadUrls,
   localStorageLanguageKey: 'language',
-  delay: 5000
+  delay: 5000,
+  landCoverLegendValues: [
```
pwambach, line 93: The main config file is not the right place for the land covers legend data. It should be provided by the layers metadata which is loaded from cloud storage

src/scripts/reducers/layers/details.ts
```diff
@@ -13,6 +15,15 @@ function detailsReducer(
 ): DetailsById {
   switch (action.type) {
     case FETCH_LAYER_SUCCESS:
+      if (action.id === 'land_cover.lccs_class') {
```
pwambach, line 18: If the layer data is in the layers metadata it would be already in the state.
And even it is not there's no need to add it here to the state. You could just import it in the component directly when the id is 'land-cover'.

pwambach (changes requested):

pwambach (approved):
nice 👍

PR #827, opened 2021-01-19
feat(filters): display filter button in data mode

### LINKED ISSUE #823: Apply tag filter to globe view
- [x] also apply the tag filter to the globe view
- [x] indicate in the UI that the current globe view is filtered
- [x] give the option to remove the filter
- [x] do not give the option to set new filters in this view

KatvonRivia:
I added a badge without the number of selected tags. It seemed a bit too small especially with two digits and was very hard to read.

image

Here is the design: https://github.com/ubilabs/esa-climate-from-space/issues/823#issuecomment-760016257

What do you think? 

closes #823

### DIFF
--- a/src/scripts/components/main/navigation/navigation.tsx
+++ b/src/scripts/components/main/navigation/navigation.tsx
@@ -13,7 +13,7 @@ import {FilterIcon} from '../icons/filter-icon';
 import setLanguageAction from '../../../actions/set-language';
 import {languageSelector} from '../../../selectors/language';
 import LanguageBubble from '../language-bubble/language-bubble';
-import TagsBubble from '../../stories/tags-bubble/tags-bubble';
+import SelectedTags from '../../stories/selected-tags/selected-tags';
 import {selectedTagsSelector} from '../../../selectors/story/selected-tags';
 import config from '../../../config/main';
 
@@ -29,7 +29,6 @@ const Navigation: FunctionComponent = () => {
     !savedLanguage
   );
   const selectedTags = useSelector(selectedTagsSelector);
-  const isFilterActive = selectedTags.length > 0;
 
   return (
     <div className={styles.navigation}>
@@ -40,7 +39,7 @@ const Navigation: FunctionComponent = () => {
         icon={StoryIcon}
         hideLabelOnMobile
       />
-      {isFilterActive && (
+      {selectedTags.length > 0 && (
         <React.Fragment>
           <Button
             className={styles.tagsButton}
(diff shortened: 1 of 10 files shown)


pmast:
Badge without number looks fine to me.

src/scripts/components/main/navigation/navigation.tsx
```diff
@@ -9,21 +9,27 @@ import {StoryIcon} from '../icons/story-icon';
...
   const dispatch = useDispatch();
   const [showMenu, setShowMenu] = useState(false);
+  const [showTags, setShowTags] = useState(false);
   const selectedLanguage = useSelector(languageSelector);
   const savedLanguage = localStorage.getItem(config.localStorageLanguageKey);
   const [showLanguageBubble, setShowLanguageBubble] = useState<boolean>(
     !savedLanguage
   );
+  const selectedTags = useSelector(selectedTagsSelector);
+  const isFilterActive = selectedTags.length > 0;
```
pwambach, line 32: Which filter do you mean?

src/scripts/components/main/navigation/navigation.tsx
```diff
@@ -34,6 +40,16 @@ const Navigation: FunctionComponent = () => {
         icon={StoryIcon}
         hideLabelOnMobile
       />
+      {isFilterActive && (
```
pwambach, line 43: I would suggest to rename `isFilterActive` to `isTagFilterActive` or you could also just say `selectedTags.length && `. This makes it easier to understand when the component should be visible.

src/scripts/components/main/navigation/navigation.tsx
```diff
@@ -65,6 +81,7 @@ const Navigation: FunctionComponent = () => {
           <Menu />
         </Overlay>
       )}
+      {isFilterActive && showTags && <TagsBubble selectedTags={selectedTags} />}
```
pwambach, line 84: The tag selector is not visible when no tag is selected? How can you select a tag then? Also TagsBubble is not a very specific name, I wasn't sure what this component does when reading it.

pwambach (changes requested):

src/scripts/components/main/navigation/navigation.tsx
KatvonRivia: The tags are selected in the stories, the user is not supposed to do that in the data view: https://github.com/ubilabs/esa-climate-from-space/issues/823

pwambach (approved):

PR #854, opened 2021-03-11
refactor(story-gallery): implement new gallery designs

### LINKED ISSUE #838: Stories: Improve image display
- [ ] add captions to side-by-side view
- [ ] add support for long captions (up to 5 lines)
- [ ] add support for headlines to captions (markdown?!)
- [ ] captions: adjust style to improve readability (size, contrast, …)
- [ ] full-screen images: take as much screen size as possible
- [ ] feedback pagination arrows are often overlooked, people miss other images (improve visibility, automatic cycling, thumbnails)
- [ ] make the captions being stored as markdown. Restrict the markdown support to headlines and links, definitely no images!

### LINKED ISSUE #840: Stories: styling
- [ ] sub headlines are missing space before

Check designs, all headlines should have space before as we cannot guarantee that they will be used in correct order.

KatvonRivia:
closes #838
closes #840


src/scripts/components/stories/story-gallery/story-gallery.tsx
```diff
@@ -3,11 +3,12 @@ import cx from 'classnames';
 
 import {PreviousIcon} from '../../main/icons/previous-icon';
 import {NextIcon} from '../../main/icons/next-icon';
-import {FullscreenExitIcon} from '../../main/icons/fullscreen-exit-icon';
 import {FullscreenIcon} from '../../main/icons/fullscreen-icon';
 import {getStoryAssetUrl} from '../../../libs/get-story-asset-urls';
 import {useInterval} from '../../../hooks/use-interval';
 import config from '../../../config/main';
+import Description from '../description/description';
```
pwambach, line 10: The name `Description` is to general here. How about `<Caption content="" />`?

pwambach (changes requested):
Screenshot 2021-03-15 at 14 47 25

The distance between the gallery dots and the caption is not the same in normal and fullscreen mode. Also the white dots are not visible on light images, can you add some drop shadow to the dots?
Do we have to use `text-align: left` here. It looks weird with short texts on larger screens, I would prefer `text-align: center`

src/scripts/components/stories/story-gallery/story-gallery.tsx
KatvonRivia: Good point, will change it.

Regarding this:
> Do we have to use text-align: left here. It looks weird with short texts on larger screens, I would prefer text-align: center

In the new designs the captions are all aligned left, but you're right it look weird with short text. I will adjust it.

KatvonRivia: image

image

KatvonRivia: I changed the background and border of the dots so you can see the progress on a white background:

image

pwambach: nice!

pwambach (approved):

PR #862, opened 2021-03-16
feat(imageType): set image types to change object fit

### LINKED ISSUE #856: add support for two image display modi:
- [ ] contain (for tables and chart)
- [ ] cover (for images to show most of the images)
- [ ] update story mapper for any newly added field

KatvonRivia:
closes #856


src/scripts/components/stories/story-gallery/story-gallery.tsx
```diff
@@ -118,7 +122,14 @@ const StoryMedia: FunctionComponent<Props> = ({
                 key={index}
                 style={{width: `${imageWidth}%`}}>
                 <div className={styles.imageContainer}>
-                  <img className={styles.photo} src={imageUrl} />
+                  <img
```
pwambach, line 125: The story gallery component is getting really big. Could you clean it up by pulling out a StoryGalleryImage component with captions for example

src/scripts/types/image-type.ts
```diff
@@ -0,0 +1,4 @@
+export enum ImageType {
+  Photo = 'photo',
+  Chart = 'chart'
```
pwambach, line 3: Chart is probably not the best type here. It could be any other image which should be contained. Why not call it what it is like 'contain' and 'cover'?

pwambach (changes requested):

src/scripts/types/image-type.ts
KatvonRivia: Sure, i can change it. I just thought contain and cover would be a bit confusing in the story mapper.

src/scripts/components/stories/story-gallery/story-gallery.tsx
KatvonRivia: aye 👌  will do

src/scripts/types/image-type.ts
pwambach: Yes good point, but I guess labeling a photograph as  "chart" just to display it as contained is more confusing. At least for me 😄

src/scripts/components/stories/story-gallery-image/story-gallery-image.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return (
+    <div
+      className={imgClasses}
+      style={{
+        width: `${containerWidth}%`,
+        transform: `translateX(-${imageWidth * currentIndex}%)`
+      }}>
+      {images.map((image, index) => {
+        const imageCaption = imageCaptions?.find((_, i) => i === index);
+        const imageType = imageFits?.find((_, i) => i === index);
```
pwambach, line 45: `const imageFit`

pwambach (changes requested):

pwambach (approved):

PR #890, opened 2021-04-13
feat(about): add route for about component

### LINKED ISSUE #886: Link top left main page log to "about CfS"
There is a wish to have a linkage to the ESA climate office page a bit more "within reach" for the users just to have it hidden in the menu. 
So the decision is to link the main page ESA logo to the About section where the user is informed about the existence of the website but still is kept in CfS application

mrMetalWood:
closes #886


pwambach (changes requested):
I would remove the 'Standalone' because when I could imagine someone who sees the component for the first time wonders if there is also a non standalone version

mrMetalWood:
Where do you still see "standalone". I renamed it to AboutProjectOverlay.

KatvonRivia (approved):

pwambach:
forget what i wrote... too early for me

pwambach (approved):

PR #917, opened 2021-05-11
fix(dataset): use correct fire legend values

pwambach:


KatvonRivia (approved):
Oh, this got lost in my mails. Sorry 🙈

PR #945, opened 2021-07-27
Feat data update

pmast:
- updated ocean colour
- updated soil moisture
- updated download for permaforst, got stuck in processing
- updated download for biomass, processing still todo


pwambach:
Why do you want to check in the tiles and these kml files? There are so many files that my browser tab freezes when I go to  the "Files changed" tab 😄

pmast:
Fixed:
image

pwambach (approved):
Nice

PR #948, opened 2021-08-03
feat(url): add language to url, replace ubilogo, close layer-selector when selected

### LINKED ISSUE #940: Layer selector: close right after selection
Currently the layer selector stays open. It has to be closed with the 'x'. Although we had arguments for that behaviour, it feels strange. especially on mobile where it covers the complete screen.
- [ ] let's try closing the drawer when a dataset is selected (we could make the behaviour different on desktop and on mobile)

### LINKED ISSUE #941: Update Ubilabs logo to new logo
We are still using the old Ubilabs Logo on the about screen.
- [ ] change the ubilabs Logo to the new logo
- [ ] change the link to ubilabs.com

### LINKED ISSUE #943: Allow to set language as URL parameter
For easier translation debugging it would be good if we could share the application with an explicitly set language.
- [ ] allow do set the language with an extra URL parameter. 
  - [ ] The URL parameter should overwrite the automatic language detection
  - [ ] the URL parameter should not overwrite the stored language for a user
  - [ ] the language chooser should behave normal and remove the URL parameter

KatvonRivia:
Append `&lng=de` or other available language to the url and refresh page to switch languages

closes #940 
closes #941 
closes #943


src/scripts/libs/language-url-parameter.ts
```diff
@@ -0,0 +1,22 @@
+import {Language} from '../types/language';
+// parses window.location and reads the story tags from query params
+//
```
pwambach, line 3: The empty line in the comment looks strange

src/scripts/libs/language-url-parameter.ts
```diff
@@ -0,0 +1,22 @@
...
+export function parseUrl(): Language | null {
+  const {hash} = location;
+
+  const queryString = hash.substr(hash.indexOf('?'));
+  const urlParams = new URLSearchParams(queryString);
+  const languageParam = urlParams.get('lng') as Language;
+
+  if (
+    (languageParam && Object.values(Language).includes(languageParam)) ||
+    !languageParam
```
pwambach, line 16: Why is there the OR statement with `|| !languageParam`? Shouldn't ir just return null if the languageParam is not included in the list?

pwambach (changes requested):

pwambach (approved):

Issue #1051, opened 2022-02-27
Show colored base map when no data is available


pmast:
- [x] Use this coilored basemap Colored Basemap: https://ubilabs.slack.com/archives/CFBKYU8D9/p1646305896145369
˜˜- [ ] Add an attribution for this dataset for 'Planetary Vision' (lets talk if this is not very easy)˜˜

pwambach:
New basemap tiles have been generated and uploaded. But I'm not sure about the attribution task. Where and why? We don't have attributions for the other basemaps

pwambach:
![Screenshot 2022-03-03 at 15 53 44](https://user-images.githubusercontent.com/1611635/156589490-0a0e48ae-3e77-470d-820d-89910fa9221f.png)

PR #1057, opened 2022-03-03
feat(basemap): add new coloured basemap

pwambach:
Generated tiles from the `satmap-v6-2-4km.tif` file

https://github.com/ubilabs/esa-climate-from-space/issues/1051


KatvonRivia (approved):
🌈

PR #1058, opened 2022-03-04
feat(onboarding): add welcome screen and onboarding tooltips

### LINKED ISSUE #1050: Add onboarding
Add onboarding.
- [ ] add a welcome screen
- [ ] allow to step through an into to the most important screen elements (TBD)
- [ ] update the cookie banner
- [ ] add a '?' to reopen the welcome screen again

KatvonRivia:
So many file changes because of translations and i had to add id's to all ui elements on the initial screen 😬  sooorry

closes #1050

To do for other PRs:
- store if onboarding was done in local storage
- hide language detection tooltip when onboarding was started


src/scripts/components/main/app/app.tsx
```diff
@@ -65,6 +71,27 @@ const TranslatedApp: FunctionComponent = () => {
         <Switch>
           <Route path="/" exact>
             {logo}
+            {showWelcomeScreen ? (
```
pwambach, line 74: I think we should abstract this more so that the onboarding and welcome stuff doesn't take up such a big part of the app component. Can you move this in a separate component which handles the state and css classes?

src/scripts/components/main/globe-navigation/globe-navigation.tsx
```diff
@@ -45,16 +45,19 @@ const GlobeNavigation: FunctionComponent<Props> = ({
     <div className={styles.globeNavigation}>
       <Button
         className={styles.projection}
+        id="projection"
```
pwambach, line 48: As the ids must be globally unique we should prefix them to avoid potential future problems. Something like `btn-projection` or `ui-projection`

src/scripts/components/main/navigation/navigation.tsx
```diff
@@ -34,6 +35,7 @@ const Navigation: FunctionComponent = () => {
     <div className={styles.navigation}>
       <Button
         className={styles.button}
+        id="stories"
```
pwambach, line 38: Same here with the ids

src/scripts/components/main/onboarding/onboarding.tsx
```diff
@@ -0,0 +1,147 @@
...
+  onPageChange: (id: number) => void;
+  onClose: () => void;
+}
+
+const Onboarding: FunctionComponent<Props> = ({id, onPageChange, onClose}) => {
+  const intl = useIntl();
+
+  const onboardingContent = [
+    {
+      id: 1,
```
pwambach, line 21: If the id is just the index it's probably better to just use the index and call the prop `step` or something. No need to manually fill in the ids here.

src/scripts/components/main/onboarding/onboarding.tsx
```diff
@@ -0,0 +1,147 @@
...
+  // set position of onboarding tooltip
+  useEffect(() => {
+    if (!referenceElement) {
+      return;
+    }
+    setReferencePosition(referenceElement.getBoundingClientRect());
+  }, [referenceElement]);
+
+  // set new position when window size changes
+  window.addEventListener(
```
pwambach, line 78: Adding an event listener has to happen in a useEffect, otherwise the event listener is added everytime the component renders.

src/scripts/components/main/onboarding/onboarding.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const referenceElement = document.querySelector(
+    `#${currentOnboarding?.elementId}`
+  );
+
+  const [referencePosition, setReferencePosition] = useState<DOMRect | null>(
+    null
+  );
+
+  // set position of onboarding tooltip
+  useEffect(() => {
```
pwambach, line 70: Why do you use the useEffect here? 
It think it should work with something like 
```
  const rect = referenceElement?.getBoundingClientRect() || null;
  const [referencePosition, setReferencePosition] = useState(rect);
```
and if it should only run when the element changes useMemo is the best hook here: 
```
  const rect = useMemo(() => referenceElement?.getBoundingClientRect() || null, [referenceElement]);
  const [referencePosition, setReferencePosition] = useState(rect);
```

src/scripts/components/main/app/app.tsx
pwambach: And perhaps integrate the overlay directly into the components?

pwambach (changes requested):

src/scripts/components/main/app/app.tsx
KatvonRivia: Good point, will change it. 👍🏼

pwambach (approved):

PR #1061, opened 2022-03-07
refactor(welcome): only show welcomeScreen once

KatvonRivia:


pwambach (changes requested):

src/scripts/libs/get-local-storage-welcome.ts
```diff
@@ -1,5 +1,11 @@
 import config from '../config/main';
 
-export default function getLocalStorageWelcomePage(): string | null {
-  return localStorage.getItem(config.localStorageWelcomePageKey);
+export default function getLocalStorageWelcomePage(): boolean {
+  const welcomeScreenChecked = localStorage.getItem(
+    config.localStorageWelcomeScreenKey
+  );
+
+  const welcomeScreenBoolean = welcomeScreenChecked === 'true' ? true : false;
```
pwambach, line 8: `? true : false` is not necessary

pwambach (changes requested):
Nice, just one tiny thing :D

pwambach (approved):

PR #1069, opened 2022-03-11
refactor(video): autoplay new video type in showcase mode

KatvonRivia:


src/scripts/components/stories/video-js/video-js.tsx
```diff
@@ -90,6 +92,8 @@ const VideoJS: FunctionComponent<Props> = ({
         ref={videoRef}
         className="video-js vjs-big-play-centered"
         style={{height: '100%'}}
+        // @ts-ignore
+        onPlay={event => onPlay(event.target)}
```
mrMetalWood, line 96: I guess this ts-ignore could be avoided like this? Or what is the issue here?
```suggestion
        onPlay={event => onPlay(event.target as VideoJsPlayer)}
```

src/scripts/components/stories/story/story.tsx
```diff
@@ -66,9 +67,15 @@ const Story: FunctionComponent = () => {
     return null;
   }
 
-  const getVideoDuration = (player: YouTubePlayer) => {
-    const duration = player.getDuration();
-    setVideoDuration(duration * 1000);
+  const getVideoDuration = (player: YouTubePlayer | VideoJsPlayer) => {
+    // @ts-ignore
```
mrMetalWood, line 71: what is the issue here?

src/scripts/components/stories/video-js/video-js.tsx
KatvonRivia: Yep, i also tried this one and EventTarget as well. 😅 
`Conversion of type 'EventTarget' to type 'VideoJsPlayer' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.`

mrMetalWood: better than the ignore I would say. This way we atleast can see what is being done and what the issue was

mrMetalWood (approved):

PR #1074, opened 2022-03-16
refactor(onboarding): highlight ui elements

KatvonRivia:
image

### DIFF
--- a/storage/stories/stories-de.json
+++ b/storage/stories/stories-de.json
@@ -1,13 +1,5 @@
 [
   {
-    "id": "debug",
-    "title": "Willkommen zu Climate from Space",
-    "description": "",
-    "image": "assets/atmospheric-ecvs.jpg",
-    "tags": [],
-    "position": [-60, -3]
-  },
-  {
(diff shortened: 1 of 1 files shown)


src/scripts/components/main/onboarding-tooltip/onboarding-tooltip.tsx
```diff
@@ -68,12 +70,20 @@ const OnboardingTooltip: FunctionComponent<Props> = ({
...
       setReferencePosition(referenceElement.getBoundingClientRect());
     };
-
     window.addEventListener('resize', setPosition);
-
     setReferencePosition(referenceElement.getBoundingClientRect());
 
-    return () => window.removeEventListener('resize', setPosition);
+    // hightlight current ui reference: move element in front of the overlay
+    referenceElement.style.zIndex = '4';
```
mrMetalWood, line 77: I would prefer top have these things attached a class in the CSS rather than inline here.

mrMetalWood (approved):

PR #1268, opened 2023-06-02
feat(globe): wait for layer to load before advancing timestep

usefulthink:


src/scripts/components/layers/time-playback/time-playback.ts
```diff
@@ -13,27 +16,60 @@ interface Props {
...
+
+  // get the state for both possible layers. When no layer is specified, we
+  //   use 'idle' as to not interfere with the loading logic
+  const layerLoadingState = useSelector(layerLoadingStateSelector);
+  const mainLayerState = mainLayerId
+    ? layerLoadingState[mainLayerId]
+    : ('idle' as LayerLoadingState);
+  const compareLayerState = compareLayerId
+    ? layerLoadingState[compareLayerId]
+    : ('idle' as LayerLoadingState);
```
usefulthink, line 59: maybe we should move this into the parent component and pass it here as a single prop `layersLoaded`? I think that feels slightly more correct than to pass the layer-ids into this component.

src/scripts/components/main/globe/globe.tsx
```diff
@@ -76,7 +81,9 @@ const Globe: FunctionComponent<Props> = props => {
 
   useProjectionSwitch(globe, projectionState.projection);
   useMultiGlobeSynchronization(globe, props);
-  // eslint-disable-next-line no-warning-comments
+
+  useLayerLoadingStateUpdater(globe, props.onLayerLoadingStateChange);
+
```
KatvonRivia, line 86: Could you pls re-add `// eslint-disable-next-line no-warning-comments` or open an issue for this? The lint-and-test run is failing. Also deployment would be blocked :)

KatvonRivia (approved):

PR #1274, opened 2023-08-09
feat(embed-wizard): implement wizard menu

KatvonRivia:
I will split the wizard in multiple parts, this PR got a bit big already - sorry 🙈 

image

in upcoming PRs:
- copy buttons, text area content
- markers, autorotate
- other settings: data-layers, stories
- language
- tags
- preview


src/scripts/components/main/embed-wizard/embed-wizard.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+import styles from './embed-wizard.module.styl';
+
+const EmbedWizard: FunctionComponent = () => {
+  const embedElements = useSelector(embedElementsSelector);
+  const [appElementsChecked, setAppElementsChecked] = useState(
+    embedElements as ElementOptions
+  );
+
+  const disabledParamsString = Object.entries(appElementsChecked)
```
pwambach, line 18: Perhaps better to pass this not as a string but as the object and only transform (with URLSearchParams) when it gets rendered

KatvonRivia: Ah of course, makes sense 👍🏼

pwambach (approved):

PR #1275, opened 2023-08-10
feat(wizard): implement previews and data layers

KatvonRivia:


src/scripts/components/main/data-viewer/data-viewer.tsx
```diff
@@ -150,37 +153,46 @@ const DataViewer: FunctionComponent<Props> = ({
...
-          (
-            {id, maxValue, minValue, units, basemap, legendValues, hideLegend},
-            index
-          ) => {
-            if (hideLegend) {
-              return null;
+      {legend &&
+        [mainLayerDetails, compareLayerDetails]
+          .filter((layer): layer is Layer => Boolean(layer))
+          .map(
```
pwambach, line 159: A bit ugly here with the destructuring :) Perhaps without it it's cleaner, or extract in a function.

pwambach (changes requested):

PR #1277, opened 2023-08-11
refactor(url): keep embed ui elements params in url

KatvonRivia:


src/scripts/components/main/url-sync/url-sync.tsx
```diff
@@ -47,11 +56,34 @@ const UrlSync: FunctionComponent = () => {
...
   }, [selectedTags, history]); // we don't want to check for location.search changes
 
+  // set embed elements query params in url when state changes
+  useEffect(() => {
+    const params = new URLSearchParams(location.search);
+
+    if (!embedParamString) {
+      embedParamOptions.forEach(option => params.delete(option));
+    } else {
+      const paramsArray = embedParamString
```
pwambach, line 66: same here with the params as string

pwambach (approved):

PR #1279, opened 2023-08-14
refactor(copy): show checkmark when copied to clipboard

KatvonRivia:


src/scripts/components/main/copy-to-clipboard-button/copy-to-clipboard-button.tsx
```diff
@@ -0,0 +1,42 @@
...
+  handleCopy: () => void;
+}
+const CopyToClipboardButton: FunctionComponent<Props> = ({
+  label,
+  handleCopy
+}) => {
+  const [copied, setCopied] = useState(false);
+
+  useEffect(() => {
+    const timeout = setTimeout(() => {
```
pwambach, line 20: Only set the timer when copied is true?

src/scripts/components/main/copy-to-clipboard-button/copy-to-clipboard-button.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+import Button from '../button/button';
+import {CopyTextIcon} from '../icons/copy-text-icon';
+import {CheckIcon} from '../icons/check-icon';
+
+import styles from './copy-to-clipboard-button.module.styl';
+
+interface Props {
+  label: string;
+  handleCopy: () => void;
```
pwambach, line 11: not sure how we handled it in this project but I thought we use 'on...' for callbacks, like `onCopy` instead of `handleCopy`

src/scripts/components/main/share/share.tsx
```diff
@@ -28,6 +29,17 @@ const Share: FunctionComponent = () => {
   const twitterUrl = replaceUrlPlaceholders(config.share.twitter, {
     currentUrl: encodeURIComponent(currentUrl)
   });
+  const [copied, setCopied] = useState(false);
+
+  useEffect(() => {
+    const timeout = setTimeout(() => {
```
pwambach, line 35: same here with the timeout

pwambach (changes requested):

src/scripts/components/main/copy-to-clipboard-button/copy-to-clipboard-button.tsx
KatvonRivia: yes, so that the checkmark only appears for a bit:
https://github.com/ubilabs/esa-climate-from-space/assets/45818654/4bf7b429-d009-4179-b235-a3d3df8a9fbd

pwambach (approved):

PR #1280, opened 2023-08-15
fix(assets): import assets directly

KatvonRivia:


src/scripts/components/main/menu/menu.tsx
```diff
@@ -1,6 +1,8 @@
 import React, {FunctionComponent, useState} from 'react';
 import {FormattedMessage} from 'react-intl';
 
+import planeratyVisionsLogo from '~/assets/images/planetary-visions.png';
```
KatvonRivia, line 4: The logos were not showing up in the deployment nor locally after we switched to vite. We are deploying them into assets not assets/images like our file structure: ` cp assets/images/* dist/assets/`.
We could copy it to an images folder inside of assets but it still won't work locally, so I'm importing them directly.

pwambach: Not sure what the problem here is, but can't we fix it? Does this occur only for the logo?

pwambach (approved):

PR #1346, opened 2024-02-14
Feat/adapt story

Immo-Be:
- add flag for new story types and configure story reducer
- adapt story 12 to show embedded iFrame chart
- fix "missing key" error in `Story.tsx`

### DIFF
--- a/src/scripts/reducers/story/selected.ts
+++ b/src/scripts/reducers/story/selected.ts
@@ -4,6 +4,7 @@ import {
 } from '../../actions/fetch-story';
 
 import {convertLegacyStory} from '../../libs/convert-legacy-story';
+import {LegacyStory} from '../../types/legacy-story';
 
 import {Story} from '../../types/story';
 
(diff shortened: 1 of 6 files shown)


src/scripts/reducers/story/selected.ts
```diff
@@ -12,8 +13,12 @@ function selectedStoryReducer(
   action: FetchStoryActions
 ): Story | null {
   switch (action.type) {
-    case FETCH_STORY_SUCCESS:
-      return convertLegacyStory(action.story);
+    case FETCH_STORY_SUCCESS: {
+      const isLegacyStory = !action.story.isCurrentStory;
```
yfr, line 17: `isCurrentStory` is used here to determine if we have a legacy story. But this prop could be set in both. Maybe change the name to make clear what this prop is meat to do. Or use some already available prop from the legacy or new story type to select which type is used.

Immo-Be: Instead of having a new property like `isCurrentStory` or `isLegacyStory` for the story, we could leave the story types as is and instead check if at least on slide of the story has `galleryItems` as property and deduct based on this if it is a "current" or "legacy" story. What do you think?

yfr: The types say that every slide in a new story should have `galleryItems`. So yes this could be a way without changing the types and adding something to the new stories that is just to mark a deprecation of another type.

Immo-Be: done :)

yfr (approved):

Issue #1349, opened 2024-02-15
Image captions should be available for iframes as well


annikabock:
We need to find a way to include the image captions we had before in the image gallery as well for an iframe

annikabock:
@Immo-Be this one was done right?

Immo-Be:
> @Immo-Be this one was done right?

Yes :)

PR #1350, opened 2024-02-16
Feat/create embedded component

Immo-Be:
also complets #1349 

- update chart with correct data
- allow description text with embedded content

image


src/scripts/components/stories/story/story.tsx
```diff
@@ -27,6 +27,7 @@ import {GalleryItemType} from '../../../types/gallery-item';
 import {useThunkDispatch} from '../../../hooks/use-thunk-dispatch';
 
 import styles from './story.module.styl';
+import StoryEmbedded from '../story-embedded/story-embedded';
```
andreashelms, line 30:
```suggestion
import StoryEmbedded from '../story-embedded/story-embedded';

import styles from './story.module.styl';
```

src/scripts/types/gallery-item.ts
```diff
@@ -26,6 +26,7 @@ export interface VideoItem {
 
 export interface EmbeddedItem {
   type: GalleryItemType.Embedded;
+  text?: string;
```
andreashelms, line 29: Maybe better `description` 🤔

Immo-Be: It was `text` before, that's we I kept it. But might be a good opportunity to use something more descriptive :)

src/scripts/components/stories/story/story.tsx
Immo-Be: 👍

src/scripts/types/gallery-item.ts
andreashelms: You are right, for other gallery types it's also text. So let's just keep that.

andreashelms (approved):

PR #1351, opened 2024-02-16
feat(globe): update to latest version with atmosphere rendering

usefulthink:
updates to the latest version of the globe which now also supports atmosphere rendering (see here for example: https://storage.googleapis.com/ubilabs-webgl-globe/examples/examples/globe-all-datasets/index.html).

image
image

### DIFF
--- a/src/scripts/components/main/globe/globe.tsx
+++ b/src/scripts/components/main/globe/globe.tsx
@@ -17,7 +17,10 @@ import {
   WebGlGlobe,
   WebGlGlobeEventMap
 } from '@ubilabs/esa-webgl-globe';
+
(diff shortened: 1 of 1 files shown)


usefulthink:
Note: build is failing because I only updated the package.json and not the package-lock.json. But I don't dare to commit that since I had to update to the new major version of vite to be able to run the app.

andreashelms:
> Note: build is failing because I only updated the package.json and not the package-lock.json. But I don't dare to commit that since I had to update to the new major version of vite to be able to run the app.

@usefulthink Should I commit that one to this branch, then?

andreashelms (approved):

PR #1359, opened 2024-02-28
feat(pipeline): vegetation lai

andreashelms:


pipeline/dags/task_factories.py
```diff
@@ -89,10 +89,10 @@ def fn(filename: str):
 def gcloud_upload_dir(layer_id: str, layer_variable: str, directory: str):
     return BashOperator(
         task_id='gcloud_upload',
-        bash_command='gcloud auth activate-service-account --key-file $KEY_FILE && gsutil -q -m cp -r $UPLOAD_DIR/* $BUCKET',
+        bash_command='gcloud auth activate-service-account --key-file $KEY_FILE && gsutil -m -h "Cache-Control:no-cache" rsync -d -r $UPLOAD_DIR $BUCKET',
```
Immo-Be, line 92: Just wondering, why the `no-cache` parameter=

Immo-Be (approved):

pipeline/dags/task_factories.py
andreashelms: You are right. I copied the command from somewhere, but in this case caching makes sense.

PR #1414, opened 2024-04-30
feat(ch4): convert csv to netcdf

yfr:


Dockerfile
andreashelms, line 1: I would probably move this and the python notebook and script to a new folder `data/csv-to-netcdf/`

yfr:
Close this because it's done

andreashelms (approved):

Issue #1402, opened 2024-04-29
chapter intro page


KatvonRivia:
chapter title page in between chapters

PR #1418, opened 2024-05-02
refactor(chapter1): tweak ui, add sections, set slideIndex in path

KatvonRivia:
closes [#1402](https://github.com/ubilabs/esa-climate-from-space/issues/1402)


andreashelms:
I think we need something like an anchor, that makes the scrolling stop at a certain point e.g. here:

image

But I don't know how this works with `react-scroll-parallax`.

andreashelms (approved):

KatvonRivia:
> I think we need something like an anchor, that makes the scrolling stop at a certain point e.g. here:
> 
> image
> But I don't know how this works with `react-scroll-parallax`.

Yeah i tried using scroll-snap-type but it didnt work 🥲

andreashelms (approved):

PR #1424, opened 2024-05-06
feat(pilot-story-globe): move globe down and in

andreashelms:
for #1412

image

### DIFF
--- a/src/scripts/components/stories/pilot-story/components/globe/globe.tsx
+++ b/src/scripts/components/stories/pilot-story/components/globe/globe.tsx
@@ -10,6 +10,7 @@ const INITIAL_DISTANCE = 30_000_000;
 
 interface Props {
   progress: number;
+  relativePosition: {top: number; left: number; right: number; bottom: number};
   isSpinning: boolean;
   isVisible: boolean;
   viewTotal: number;
@@ -18,6 +19,7 @@ interface Props {
 
 const Globe: FunctionComponent<Props> = ({
   progress,
+  relativePosition: initialPosition,
   isSpinning,
   isVisible,
   viewTotal,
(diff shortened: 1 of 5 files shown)


src/scripts/components/stories/pilot-story/components/globe/globe.tsx
```diff
@@ -63,6 +65,16 @@ const Globe: FunctionComponent<Props> = ({
...
   }, [containerRef, globe]);
 
+  useEffect(() => {
+    if (containerRef.current) {
+      containerRef.current.style.top = `${initialPosition.top}%`;
+      containerRef.current.style.left = `${initialPosition.left}%`;
+      containerRef.current.style.right = `${initialPosition.right}%`;
+      containerRef.current.style.bottom = `${initialPosition.bottom}%`;
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Immo-Be, line 75: You should be able to memoize `initialPosition` and get rid of the comment here

Immo-Be (approved):

src/scripts/components/stories/pilot-story/components/globe/globe.tsx
andreashelms: Can you please explain how?

Immo-Be: As `initialPosition` is passed down as a prop, you could define it in the parent like 

```
  const memoizedRelativePosition = useMemo(
    () => ({top: 0, left: -100, right: 0, bottom: 0}),
    []
  );
```

andreashelms: Ok, will be removed with next PR.

PR #1425, opened 2024-05-07
feat(pilot-story-globe): set relative extend to movement

andreashelms:


src/scripts/components/stories/pilot-story/pilot-story.tsx
```diff
@@ -24,24 +24,34 @@ const PilotStory: FunctionComponent = () => {
...
-          viewTotal={5}
+          pagesTotal={5}
           globeMovement={[
             {
-              viewFrom: 1,
-              viewTo: 2,
-              directions: [GlobeMovementDirection.RIGHT]
+              pageFrom: 1,
+              pageTo: 2,
+              relativeExtend: 30,
```
andreashelms, line 35: sets the extend for a movement. This means globe is moved by 30% right. Relative x position was set to -30,  so globe is centered after movement is finished on page 2.

Immo-Be (approved):

PR #1427, opened 2024-05-08
feat(carouse): implement carousel

KatvonRivia:


andreashelms (approved):

src/scripts/components/stories/pilot-story/components/carousel/carousel.tsx
Immo-Be, line 1: As we have different carousels in the app, could you find a name that is specific to this type of carousel?

KatvonRivia: Oh yes, forgot to clean that up

PR #1435, opened 2024-05-13
refactor(pages): adjust scroll snap and positioning

KatvonRivia:


src/scripts/components/stories/pilot-story/components/pilot-carousel/pilot-carousel.tsx
```diff
@@ -56,6 +56,13 @@ const PilotCarousel = () => {
 
   return (
     <div className={styles.carouselWrapper}>
+      <h1>Some of the primary sources include:</h1>
```
KatvonRivia, line 59: @Immo-Be i moved the legend inside the carousel so its within the 100vh for scroll snapping

PR #1497, opened 2024-05-22
feat(chapter-markers): floating button for next chapter

andreashelms:
image

### DIFF
--- a/src/scripts/components/stories/pilot-story/components/06-chapter/06-chapter.tsx
+++ b/src/scripts/components/stories/pilot-story/components/06-chapter/06-chapter.tsx
@@ -46,7 +46,6 @@ export interface SubStoryContent {
 const ChapterSix: FunctionComponent<Props> = ({chapterIndex}) => {
   const history = useHistory();
   const location = useLocation();
-  const {setSelectedChapterIndex} = useChapter();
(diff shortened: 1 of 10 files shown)


src/scripts/components/stories/pilot-story/components/06-chapter/06-chapter.tsx
```diff
@@ -82,6 +88,11 @@ const ChapterSix: FunctionComponent<Props> = ({chapterIndex}) => {
     storyRef.current?.scrollIntoView({behavior: 'instant'});
   };
 
+  const handleChapterChange = () => {
+    setSelectedChapterIndex(6);
+    scrollToChapterIndex(6);
```
andreashelms, line 93: @Immo-Be You are still working on this, right? Because the change to chapter 7 does not work here.

KatvonRivia (approved):

src/scripts/components/stories/pilot-story/components/06-chapter/06-chapter.tsx
Immo-Be: Indeed :)

Immo-Be (approved):

PR #1652, opened 2025-02-05
Chore/update dev deps

Immo-Be:
for #deps


src/electron/main.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
 import * as path from 'path';
 import {app, BrowserWindow, ipcMain} from 'electron';
 const isDev = require('electron-is-dev');
```
andreashelms, line 4:
```suggestion
import * as path from 'path';
import {app, BrowserWindow, ipcMain} from 'electron';
// eslint-disable-next-line @typescript-eslint/no-require-imports
const isDev = require('electron-is-dev');
```

eslint.config.js
```diff
@@ -0,0 +1,30 @@
...
+    },
+    rules: {
+      ...reactHooks.configs.recommended.rules,
+      'react-refresh/only-export-components': [
+        'warn',
+        {allowConstantExport: true}
+      ]
+    }
+  }
+);
```
andreashelms, line 30: Where does this config come from?

Immo-Be: It's the default config that is generated when you setup a project with vite + react + ts today

andreashelms:
Thanks! 💪

andreashelms (approved):

PR #1718, opened 2025-03-19
feat: enable compare mode of globes

Immo-Be:


Immo-Be:
closed in favour of #1724

PR #1740, opened 2025-03-24
fix(marker-markup): decrease marker svg size

### LINKED ISSUE #1683: marker seems in the athmosphere
Image

andreashelms:
closes #1683

Before
image
Now
image

### DIFF
--- a/src/scripts/components/main/globe/marker-markup.tsx
+++ b/src/scripts/components/main/globe/marker-markup.tsx
@@ -13,13 +13,19 @@ export const MarkerMarkup: FunctionComponent<Props> = ({ title }) => {
       className={styles.marker}
       data-marker={title}
       xmlns="http://www.w3.org/2000/svg"
-      width="48"
-      height="48"
-      viewBox="0 0 48 48"
+      width="32"
+      height="32"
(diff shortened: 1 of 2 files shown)


src/scripts/components/main/globe/marker-markup.tsx
```diff
@@ -13,13 +13,19 @@ export const MarkerMarkup: FunctionComponent<Props> = ({ title }) => {
       className={styles.marker}
       data-marker={title}
       xmlns="http://www.w3.org/2000/svg"
-      width="48"
-      height="48"
-      viewBox="0 0 48 48"
+      width="32"
+      height="32"
+      viewBox="0 0 32 32"
```
Immo-Be, line 18: This size was intentional, because they are too small on desktop. As this is mainly a mobile problem, maybe we can have different sizes for mobile and dekstop?

andreashelms: Actually the marker is not smaller when being displayed. I changed marker size to the size of the former marker to be displayed at the correct position.

Immo-Be (approved):

Issue #408, opened 2020-07-22
Datasets: Permafrost


pmast:
- [ ] use the colormap from PE:
```
100 37 87 172 255
0 37 87 172 0
```

pwambach:
This is the one we had in the beginning. gdal didn't use the correct transparency values here so everything was either blue or completely transparent. The one we use now has just more steps in between

```
100 37 87 172 255
80 37 87 172 204
60 37 87 172 153
40 37 87 172 102
20 37 87 172 51
0 37 87 172 0
nv 0 0 0 0
```

Is something wrong with how it looks right now?

pmast:
Looks much better! I will send this for review, then we will see!

pmast:
Updated data set: 
Permafrost: /neodc/esacci/permafrost/data/permafrost_extent/L4/area4/pp/v02.0

pwambach:
V02.0 has some strange behaviour. I tried out a lot of projection stuff but nothing works. 

gdalinfo outputs:
```
polar_stereographic#crs_wkt=PROJCS["WGS 84 / Arctic Polar Stereographic",GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AUTHORITY["EPSG","4326"]],PROJECTION["Polar_Stereographic"],PARAMETER["latitude_of_origin",71],PARAMETER["central_meridian",0],PARAMETER["scale_factor",1],PARAMETER["false_easting",0],PARAMETER["false_northing",0],UNIT["metre",1,AUTHORITY["EPSG","9001"]],AXIS["X",EAST],AXIS["Y",NORTH],AUTHORITY["EPSG","3995"]]
```

but using `gdalwarp -s_srs EPSG:3995` doesn't fit

pmast:
Version 3.0 is available: https://catalogue.ceda.ac.uk/uuid/6e2091cb0c8b4106921b63cd5357c97c

pmast:
Tried to generate tiles. 

- downloaded data files succesfuly
- skipped the `add-time-coordinate.py` part of the download script as the data already has a time dimension
- uploaded the data to storage
- run the trigger
- meta data got updated and seems to be correct
[...]

PR #1768, opened 2025-04-23
feat(ci): add docker container for toolbox

andreashelms:
for #408


ci/cloudbuild-build-toolbox.yaml
```diff
@@ -0,0 +1,16 @@
+# build with "gcloud builds submit --config ./ci/cloudbuild-build-toolbox.yaml --timeout=2000 ./ci/docker"
+
+
+steps:
+  - name: 'gcr.io/cloud-builders/docker'
```
Immo-Be, line 5: Suggestions
Ensure that $PROJECT_ID is set correctly in the environment where this script runs.
Consider adding validation steps or tests to the build (e.g., validating that the image works as intended after being built).
If this is a production-critical build, you might want to define a timeout more conservatively than 2000 seconds.

Immo-Be (approved):

PR #1794, opened 2025-06-05
fix(layers): layers selector language parameter

andreashelms:


src/scripts/selectors/layers/list.ts
```diff
@@ -2,8 +2,18 @@ import { State } from "../../reducers/index";
 import { layersApi } from "../../services/api";
 import { createSelector } from "@reduxjs/toolkit";
 
+// Get current language from the state
+const selectCurrentLanguage = (state: State) => state.language;
+
 // Get layers list
-export const selectLayers = layersApi.endpoints.getLayerList.select("en");
+export const selectLayers = createSelector(
```
Immo-Be, line 9: good catch...

Immo-Be (approved):

PR #1831, opened 2025-06-25
fix(embed-wizard): update params

### LINKED ISSUE #1820: update embed wizard
Image

@annikabock 
What feature parameters should the new embed wizard support?

Immo-Be:
closes #1820. 

This updates the embed wizard options to the new navigation. It also improves embed wizard logic and UX

With header enabled:
image

With header disabled:
image

### DIFF
--- a/src/scripts/components/main/overlay/overlay.tsx
+++ b/src/scripts/components/main/overlay/overlay.tsx
@@ -23,8 +23,20 @@ const Overlay: FunctionComponent<Props> = ({
   const modalElement = document.getElementById("modal");
   const classes = cx(styles.overlay, className);
 
+  // stops progation when overlay is present
+  // This is useful because otherwise the user would also
+  // control the gesture-based navigation and interfere with the overlay
+  const stopPropagation = (e: React.SyntheticEvent) => {
+    e.stopPropagation();
+  };
+
   const Content = (
-    <div className={classes}>
+    <div
+      className={classes}
+      onWheel={stopPropagation}
+      onMouseDown={stopPropagation}
+      onTouchStart={stopPropagation}
+    >
       {showCloseButton && (
         <Button
           icon={CloseIcon}


src/scripts/components/main/navigation/navigation.tsx
```diff
@@ -2,26 +2,32 @@ import { FunctionComponent, useState } from "react";
...
-import { useContentParams } from "../../../hooks/use-content-params";
-import useIsStoriesPath from "../../../hooks/use-is-stories-path";
-import { useScreenSize } from "../../../hooks/use-screen-size";
-import { EsaLogo } from "../icons/esa-logo";
-import { ArrowBackIcon } from "../icons/arrow-back-icon";
-import { setShowLayer } from "../../../reducers/show-layer-selector";
-import { StoryMode } from "../../../types/story-mode";
-import { LayerSelectorIcon } from "../icons/layer-selector-icon";
-import { contentSelector } from "../../../selectors/content";
+import { embedElementsSelector } from "../../../selectors/embed-elements-selector";
```
andreashelms, line 30: Please sort imports

src/scripts/components/main/navigation/navigation.tsx
```diff
@@ -31,18 +37,26 @@ const Navigation: FunctionComponent = () => {
...
 
-  const {category} = useSelector(contentSelector)
+  const { category } = useSelector(contentSelector);
   const isStoriesPath = useIsStoriesPath();
 
   const { isNavigation } = useContentParams();
   const { isMobile } = useScreenSize();
 
   const { mode } = useContentParams();
+  const { header, logo, back_link, app_menu, layers_menu } = useSelector(
```
andreashelms, line 47: I'd prefer camel case here. Maybe add show prefix. e.g. `showHeader`

src/scripts/libs/create-embed-url.ts
```diff
@@ -1,11 +1,34 @@
...
+    uncheckedKeys.forEach((key) => {
+      params.set(key, "false");
+    });
+  }
+
+  // Rebuild the new hash part
+  const newHash = `${hashPath}${params.toString() ? `?${params.toString()}` : ""}`;
+
+  // Construct and return full updated URL
+  return `${origin}${pathname}${newHash}`;
```
andreashelms, line 33:
```suggestion
  return origin + pathname + newHash;
```

src/scripts/components/main/navigation/navigation.tsx
Immo-Be: I would keep it this way, because they are also the params in the url where snake case seems appropriate. Also, it has already been implemented like this throughout the app.

andreashelms:
If back arrow is hidden layers menu is moved to the left.

image

Do we need all these settings at all?
Wouldn't it be sufficient to make header, time-slider and legend hideable?

andreashelms:
@Immo-Be Wizard is not scrollable. I can not see the options at the bottom.

image

Immo-Be:
> Just noticed, embed wizard looks broken on mobile 😕
> 
> I know, but it was always like that. As this PR is quite big, let's merge this and create new issue, ok?

andreashelms (approved):

Issue #1839, opened 2025-07-02
scroll concept evaluation


Immo-Be:
we need to decide how to implement the scrolling gesture on mobile and desktop for story content

- consider learnings from pilot story

Goals: 
Modularity and Flexibility
Easy-to-use API
Synchronization with all UI elements

Idea:
```
<Stories> (custom Parallex Context Provider)
	<Story>
		<ImageGallery> (root compound component)
			<ImageGallery.FrequencyBlend/>
			<ImageGallery.CompareMode/>
			<ImageGallery.TimeBlend/>
			<ImageGallery.ScrollCaption/>
			<ImageGallery.ScrollOverlay/>
		</ImageGallery>
	</Story>
	<Story>
		<TextBlock> (root compound component)
				…
		</TextBlock>
	</Story>
</Stories>
```

PR #1853, opened 2025-07-03
Feat/story redesign implementation

Immo-Be:
for #1839 

- move 'legacy' story files a separate folder
- rename some "navigation.tsx" to "header.tsx" (because that's what is it)


andreashelms:
Do we need to create a new branch for story changes? I am not sure wether this should be on the next release or the release after that.

PR #1866, opened 2025-07-22
Feat/gallery background

Immo-Be:
closes #1864 

Screenshot 2025-07-22 at 18 14 44


src/scripts/components/stories/story/blocks/image-gallery/formats/scroll-overlay/scroll-overlay-slide/scroll-overlay-slide.module.css
```diff
@@ -0,0 +1,72 @@
...
+    will-change: transform;
+  }
+
+  .asset {
+    width: 100%;
+    height: 100%;
+    object-fit: cover;
+  }
+
+  &:not(.first):before {
```
yfr, line 27: There is a selector for the first. https://developer.mozilla.org/en-US/docs/Web/CSS/:first-child

PR #1874, opened 2025-08-04
feat(globe): implement globe

Immo-Be:


package.json
```diff
@@ -47,7 +47,7 @@
   "dependencies": {
     "@datapunt/matomo-tracker-react": "^0.5.1",
     "@reduxjs/toolkit": "^2.5.1",
-    "@ubilabs/esa-webgl-globe": "^1.2.0",
```
Immo-Be, line 50: will make a release once approved

src/scripts/hooks/use-globe-route-state.ts
```diff
@@ -65,15 +81,14 @@ export function useGlobeRouteState() {
...
-          dispatch(setIsAutoRotating(true));
           break;
         }
 
-        // Reset the globe view to the default state
         dispatch(
           setFlyTo({
             ...config.globe.view,
             isAnimated: true,
+            interpolationFactor: 0.08,
```
andreashelms, line 91: What is that?

Immo-Be: The interpolationFactor essentially replaces `animationDuration` in the way that it determines how fast the globe `flyTo` is performed

Issue #104, opened 2019-09-25
accessibility


JuJuPlath:
- focus style
- tab order
- set aria-labels for buttons/links without texts

PR #1888, opened 2025-08-12
feat(story): add name and aria-label to link and button

Immo-Be:
for #104


andreashelms (approved):

Immo-Be:
@andreashelms I added a package `eslint-plugin-jsx-a11y` which marks accessibility problems via eslint rules. I also implemented fixes here. Please re-review

Issue #1942, opened 2025-09-03
bugs on downloaded version


annikabock:
- [x] 1. I keep getting a javascript error – persistent after a couple of clicks. Can you advise.
- [x] 2. Downloading missing datasets – need to downloaded on startup. On opening the app most datasets are available. Several need to be downloaded – which was expected and obviously needs a wifi connection but when the app is restarted the data needs to be downloaded again, ie it deosnt appear to be held locally on the machine so the process needs repeating. This is awkward and appears to be a new problem compared to when I’ve done this before with earlier iterations ie I previously downladed the data once and it worked.
eg

in Fr:

Biomasse en surface - the globe is free of any previously available data

Prender le opuls de la planete - this hadn't been accessed before but when opened with no wifi the screen is blank.

 

This isn't really practical as it would mean the person managing the app from the ESA side would have to get a a wifi connection, open everything I reckon to get a download.

- [ ] 3. If wifi turns off the animations don’t work. This sort a significant problem as the standalone version is for use in standalone situations eg exhibitions where no wifi or reliable wifi is available. Can you investigate. The previous iterations of the standalone version didn’t have this issue.

PR #1944, opened 2025-09-03
fix(electron): ipc handlers registered multiple times

andreashelms:
for #1942


Immo-Be:
Did you check the gemini comment?

Issue #2029, opened 2025-12-11
Update React to v19.2


andreashelms:
this would be a great case to use `useEffectEvent` hook, but I don't think we are on React v19.2 in this project? I wouldn't mind if we update now but could also refactor later

_Originally posted by @Immo-Be in https://github.com/ubilabs/esa-climate-from-space/pull/2024#discussion_r2610073753_

PR #2045, opened 2025-12-18
chore: update dependencies

Immo-Be:
update deps, including React `v19.2.3` EXCLUDING all deps related to `Electron`

for #2029

### DIFF
--- a/storage/stories/debug-legacy/debug-legacy-de.json
+++ b/storage/stories/debug-legacy/debug-legacy-de.json
@@ -1,9 +1,9 @@
 {
-  "id": "debug",
+  "id": "debug-legacy",
   "slides": [
     {
       "type": "splashscreen",
-      "text": "# ENGLISH Is Ozone Good or Bad?\n\nThe ozone layer protects life on Earth from ultraviolet solar radiation, but ozone is also a powerful greenhouse gas and at ground level is extremely hazardous to health.",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -47,68 +47,68 @@
...
     "lodash.intersection": "^4.4.0",
-    "motion": "^12.23.3",
-    "react": "^19.0.0",
-    "react-dom": "^19.0.0",
-    "react-intl": "^7.1.5",
-    "react-markdown": "^8.0.3",
+    "motion": "^12.23.26",
+    "react": "^19.2.3",
+    "react-dom": "^19.2.3",
+    "react-intl": "^8.0.4",
```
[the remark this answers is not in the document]
Immo-Be: Everything worked fine when i tested but we should keep an open eye on any bugs for the next developing cycle...

src/scripts/components/legacy-stories/layer-description/layer-description.tsx
```diff
@@ -13,7 +14,7 @@ const LayerDescription: FunctionComponent<Props> = ({ layerDescription }) => (
   <div className={styles.layerDescription}>
     <ReactMarkdown
       children={layerDescription}
-      linkTarget="_blank"
+      rehypePlugins={[rehypeRaw]}
```
andreashelms: Can you please explain why this is needed?

andreashelms: Ok, I got it.

PR #2046, opened 2025-12-18
chore: add missing translations

Immo-Be:


andreashelms:
Maybe this PR is worth to be a hotfix, too...

PR #2060, opened 2026-01-26
feat(image-carousel): add new module based on image scroll

andreashelms:
image
image


Immo-Be:
> i get different image sizes

i think that might only happen when an image source ist not found

Immo-Be:
i get weird overflow when tabbing through the caroussel. Can you check?

Screenshot 2026-01-28 at 15 38 51

PR #2064, opened 2026-02-03
feat(story-globe): add story globe to module based stories

andreashelms:
image
image

### DIFF
--- a/src/scripts/components/stories/story/blocks/globe/story-globe/story-globe.tsx
+++ b/src/scripts/components/stories/story/blocks/globe/story-globe/story-globe.tsx
@@ -60,7 +60,7 @@ const StoryGlobe: FunctionComponent = () => {
    * @param layerIds - Array of layer IDs to retrieve names for
    * @returns Formatted string of layer names or undefined if no valid names found
    */
-  const getLayerNames = (layerIds: string[] | undefined) => {
+  const formatLayerNames = (layerIds: string[] | undefined) => {
     if (!layerIds || !layers) return "";
     return layerIds
       .map((id) => layers.find((layer) => layer.id === id)?.name)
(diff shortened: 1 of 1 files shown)


Immo-Be:
I tested the globe implementation and looks good already :) some stuff i have noticed:
- Globe flashes in the background when story is loaded (we had that before, but now we briefly see the component with text, and legends etc.)
- Tabbing out of the component does not deactivate “interaction mode” ➡︎ page seems frozen 
- Interacting with the globe, updates the globe wrongfully: We store the current “page”  as a parameter behind the story e.g. `debug/17`. When I start to interact, the url is set to `debug/0`. This would then cause problems when the user shares the url, changes language etc. 
- When I have more than one globe component, we have two problems: 1. I always get the same layer, (presumably the layer of the last globe), even when there are different layers specified. 2. I seem to get a “giant globe” on all globe instances, which seems to persist through reload (encoding in url)

Screenshot 2026-02-05 at 11 17 45

src/scripts/components/stories/story/blocks/globe/story-globe/story-globe.tsx
```diff
@@ -0,0 +1,108 @@
...
+          aria-label={intl.formatMessage({ id: "storyGlobe.quitInteraction" })}
+        ></Button>
+      )}
+      <Globe
+        touchable={isInteractive}
+        className={styles.globeContainer}
+        globeItem={{
+          ...globe,
+          layerDescription: getLayerNames(
+            globe.layer?.map((layer) => layer.id),
```
Immo-Be: a `globe.layer` has multiple layers? 🤔

src/scripts/components/layers/time-slider/time-slider.tsx
```diff
@@ -39,11 +42,19 @@ const DELAY = 200;
...
 }) => {
   const selectedLayerIds = useSelector(selectedLayerIdsSelector);
-  const { mainId, compareId } = selectedLayerIds;
+  let { mainId, compareId } = selectedLayerIds;
+  if (mainLayerId) {
+    mainId = mainLayerId;
+  }
+  if (compareLayerId) {
+    compareId = compareLayerId;
+  }
```
Immo-Be: this is not clear to me. `mainId` and `compareId` is from the the global state, and here you override them with another id? This needs clarification or can we do something like
`const newMainId = mainId ?? mainLayerId`

PR #2105, opened 2026-02-26
feat(story): set water mask layer for eei story

andreashelms:
for #2092

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,7 +69,7 @@
         "postcss": "^8.5.6",
         "postcss-custom-media": "^11.0.6",
         "prettier": "3.7.4",
-        "rollup": "^4.53.5",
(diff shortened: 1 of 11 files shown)


src/scripts/components/stories/story/story.tsx
```diff
@@ -27,6 +29,16 @@ import styles from "./story.module.css";
 const Story: FunctionComponent = () => {
   const { storyElementRef, story, setScrollAnchorRefs } = useStory();
   const { isStoryEEI } = useAppRouteFlags();
+  const dispatch = useDispatch();
+
+  useEffect(() => {
```
Immo-Be: cool! let's keep `story.tsx` clean of anything too specific though. Can you move this do another component / hook? e.g. `useGlobeRouteState`

Issue #2221, opened 2026-03-19
Embedded globe needs title and caption


annikabock:
The title will come from the spreadsheet and we need space for a caption

- [x] update story mapper
- [ ] update globe module

PR #2241, opened 2026-03-23
feat(story-globe): implement title and caption

Immo-Be:
for #2221


src/scripts/components/layers/time-slider/time-slider.tsx
```diff
@@ -33,6 +34,7 @@ interface Props {
   setGlobeTime: (time: number) => void;
   noTimeClamp?: boolean;
   autoplay?: boolean;
+  customStyles?: CSSProperties;
```
yfr: Can't you just use `styles` for the name of the property? Would be much more in line with standart naming convention.

Immo-Be: i am already using `styles` for the module import...

PR #2316, opened 2026-05-12
Chore/story eei feedback

### LINKED ISSUE #2311: Satellite when moves in and out
If you are not constantly scrolling the moving satellite is already disappeard when it comes to the text regarding the satellite. So the solution should be:

Move in the satellite when the text bock regarding the satellite appears. Fix the Satellite in the upper left corner (I think there is a design for that somewhere before we decided to make it moving) and keep it a loooong time on the same place..... move it out to the right side before the screen with the counter of kettles boiled appears

### LINKED ISSUE #2312: font size of time slider
it was critisised that the font size of the time slider in desktop view is to fiddly. Please make it bigger. Think the next font size we have is the size the copy in the box has. Everyone is fine to take that one

### LINKED ISSUE #2313: adjust scrolling behaviour
Scrolling Speed/Behaviour

- Fewer scrolls to move on from the first scene would be better, it takes ~4 scrolls to get from “More energy comes in…” to “Increasing amounts…” and another ~3 to zoom into the globe.

- Having a scrollbar to keep track of navigation has been mentioned. You previously clarified the current approach is to wait to see what format and features work best across all the modules once they are in place; just wanted to highlight this idea again.

- Straight after this is a globe with nothing else, feels like the end; could do with a faster transition/next text coming in sooner.

Immo-Be:
closes #2311 
closes #2312
closes #2313

➡︎ there seems to be a problem where some of the animations were not applied correctly anymore. I updated the package-lock to resolve that


src/scripts/components/stories/story/blocks/story-eei/satellite-animation/satellite-animation.tsx
```diff
@@ -0,0 +1,74 @@
...
+    // make sure animation restarts
+    if (latest < config.xPosition.scrollReset) {
+      animate(
+        scope.current,
+        { x: config.xPosition.initial, y: "0px" },
+        { duration: 0 },
+      );
+    }
+
+    if (quantize(latest, 0.01) === config.xPosition.scrollStart) {
```
andreashelms, line 57: What does this do? Maybe add a comment

andreashelms (approved):

PR #2325, opened 2026-05-20
feat(content-navigation): add scroll hints for datasets and stories

### LINKED ISSUE #2308: additional design needed for this
- [x] we need a toggle in the view where data layers are compared, to decouple the syncronisation of the globes. See #2306 
- [x] please add a little indicator like the "scroll to navigate" indicator in the front page that helps to understand that you scoll up for stories and down for data layers. I know there is little space on mobile but maybe you can find a way?

andreashelms:
closes #2308 

image
image


src/scripts/components/main/content-navigation/content-navigation.module.css
```diff
@@ -208,3 +208,82 @@
     }
   }
 }
+
```
plumdumpling, line 211: To improve scalability and accessibility, using rem / em over px for font size and spacing (gap, padding, line-height, letter-spacing, etc.) would be better. Though I'm not sure if it fits the rest of the project's styles.

plumdumpling (approved):

Issue #2315, opened 2026-05-12
some text adjustments


annikabock:
coming soon

annikabock:
[CCI-EarthEnergyImbalance-storyboard-v4.docx](https://github.com/user-attachments/files/27963844/CCI-EarthEnergyImbalance-storyboard-v4.docx)

annikabock:
some weird positioning of the texts :) 
Would be good to have

it like:
_Above the earth_
Although the ocean covers 71% of Earth’s surface,...

_above the square_
... it absorbs 91% of the incoming energy.

and then accordingly for all following parts.
oh and Sorry, we need to switch earth and square for desktop so it makes sense then with the sentence. in mobile it is correctly distributed

<img width="1899" height="763" alt="Image" src="https://github.com/user-attachments/assets/c81951f3-56ff-46d8-9763-bfc64e3d8d21" />

Oh and here - missing in Philips corrections - there should be written

- MOTECOSUMA Project Team (with the dash in front)

<img width="396" height="143" alt="Image" src="https://github.com/user-attachments/assets/f3251e7f-1e8a-4218-9d98-c09f71552f89" />


and :) you did not erase the "nightlight" in the text

<img width="692" height="235" alt="Image" src="https://github.com/user-attachments/assets/ec383e75-bf15-47ad-98e3-48aac536386e" />

PR #2333, opened 2026-05-21
feat(story-eei): module -and text adjustments

Immo-Be:
#2315


src/scripts/components/stories/story/blocks/story-eei/modules/tree-map/tree-map-grid/tree-map-grid.module.css
```diff
@@ -1,15 +1,10 @@
 .gridContainer {
-  width: 90%;
-  margin-top: 25px;
+  width: 100%;
+  margin-block: 25px;
```
KatvonRivia, line 3: use rem? :)

KatvonRivia (approved):
+ the hardcoded strings in english gemini mentioned (if not on purpose)

PR #2365, opened 2026-06-11
fix(route): correctly route legacy url to data view

Immo-Be:
- correctly routes legacy routes embedding data sets from esa website, e.g.
  https://climate.esa.int/en/projects/land-surface-temperature/
  - restores url param `stories_menu`


src/scripts/hooks/use-globe-route-state.ts
```diff
@@ -39,6 +40,28 @@ export function useGlobeRouteState(globe: WebGlGlobe | null) {
   const language = useSelector(languageSelector);
   const { data: layers } = useGetLayerListQuery(language);
 
+  const redirectLegacyLayerUrl = useCallback(() => {
```
andreashelms, line 47: Add a comment above, explaining what is happening here

andreashelms (approved):

PR #2376, opened 2026-06-18
feat(story-embedded): make embedded element responsive to container

andreashelms:
image

### DIFF
--- a/src/scripts/components/legacy-stories/caption/caption.tsx
+++ b/src/scripts/components/legacy-stories/caption/caption.tsx
@@ -1,33 +1,27 @@
 import { FunctionComponent } from "react";
 import ReactMarkdown from "react-markdown";
 import cx from "classnames";
-import { useScreenInfo } from "../../../hooks/use-screen-info";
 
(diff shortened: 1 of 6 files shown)


src/scripts/components/legacy-stories/caption/caption.tsx
```diff
@@ -1,33 +1,27 @@
...
 import cx from "classnames";
-import { useScreenInfo } from "../../../hooks/use-screen-info";
 
 import styles from "./caption.module.css";
-import { ImageFit } from "../../../types/image-fit";
 
 interface Props {
   content: string;
   showLightbox: boolean;
-  imageFit?: ImageFit;
```
Immo-Be, line 12: not sure if this make break stuff?

Immo-Be (approved):

src/scripts/components/legacy-stories/caption/caption.tsx
andreashelms: Yes, it is not used anywhere else than in `story-image` and `story-embedded`.

PR #2390, opened 2026-07-21
add chapter progress indicator for stories

### LINKED ISSUE #2228: chapter progress indicator

Immo-Be:
closes #2228 

- it differs slightly from the original intent where each slide was represented by a dot. Instead, we have a dot per module and the section's length of the "indicator line" represent each module's length. We do this, because for longer stories (especially on mobile), the dots would be really tightly placed to each other

Screenshot 2026-07-30 at 17 10 04

### DIFF
--- a/src/scripts/components/main/chapter-indicator/chapter-indicator.tsx
+++ b/src/scripts/components/main/chapter-indicator/chapter-indicator.tsx
@@ -1,4 +1,5 @@
-import { CSSProperties } from "react";
+import { CSSProperties, useState } from "react";
+import { motion } from "motion/react";
 import styles from "./chapter-indicator.module.css";
 
 interface Props {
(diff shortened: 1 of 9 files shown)


src/scripts/components/main/chapter-indicator/chapter-indicator.tsx
```diff
@@ -0,0 +1,103 @@
...
+          style={{
+            positionAnchor: `--dot-${activeIndicator}`,
+          }}
+        ></motion.span>
+      </ol>
+    </motion.nav>
+  );
+};
+
+export default ChapterIndicator;
```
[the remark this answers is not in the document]
Immo-Be: we don't use `anchor-positioning` anymore

src/scripts/hooks/use-module-scroll.ts
```diff
@@ -0,0 +1,45 @@
...
+
+      return heigth;
+    });
+
+    setHeightFractionPerModule(scrollYPerModule);
+  });
+
+  useEffect(() => {
+    // eslint-disable-next-line react-hooks/set-state-in-effect
+    onMount();
```
Immo-Be, line 36: not really sure how to avoid this here because we can only use the nodes for calculation AFTER they have been rendered...

src/scripts/components/stories/story/blocks/splashscreen/splashscreen.tsx
```diff
@@ -25,7 +25,9 @@ import cx from "classnames";
 
 import styles from "./splashscreen.module.css";
 
-export const SplashScreen: FunctionComponent<StorySectionProps> = () => {
+export const SplashScreen: FunctionComponent<StorySectionProps> = ({
+  ...rest
+}) => {
```
plumdumpling: No spread needed here
```suggestion
export const SplashScreen: FunctionComponent<StorySectionProps> = (rest) => {
```

src/scripts/components/stories/story/story.tsx
```diff
@@ -45,7 +47,8 @@ const Story: FunctionComponent<{ children?: ReactNode }> = ({ children }) => {
         ref={storyElementRef}
         id="story"
       >
-        <SplashScreen />
+        <ChapterIndicator />
+        <SplashScreen ref={setModuleRefs(`${0}`)} />
```
plumdumpling:
```suggestion
        <SplashScreen ref={setModuleRefs('0')} />
```

src/scripts/hooks/use-module-scroll.ts
plumdumpling: I think it's ok, but you could use `useLayoutEffect` here to avoid an unnecessary paint.

plumdumpling (approved):

src/scripts/hooks/use-module-scroll.ts
Immo-Be: > I think it's ok, but you could use `useLayoutEffect` here to avoid an unnecessary paint.

`useLayoutEffect` causes  the `chapter-indicator` not to render probably. I guess I just keep it this way

PR #2401, opened 2026-07-31
Feat/setup xFires story

plumdumpling:
- set up new xFires story

### DIFF
--- a/src/scripts/components/stories/story/scroll-story/story-x-fires/splashscreen/splashscreen-x-fires.tsx
+++ b/src/scripts/components/stories/story/scroll-story/story-x-fires/splashscreen/splashscreen-x-fires.tsx
@@ -1,0 +1,106 @@
+import { FunctionComponent } from "react";
+import { motion, useTransform } from "motion/react";
+
+import { useScreenInfo } from "../../../../../../hooks/use-screen-info";
+import { useScrollModule } from "../../modules/base-scroll/use-scroll-module";
+import { useIntl } from "react-intl";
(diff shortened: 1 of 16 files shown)


src/scripts/components/stories/story/scroll-story/story-x-fires/splashscreen/splashscreen-x-fires.tsx
```diff
@@ -0,0 +1,106 @@
...
+          animationConfig.gestureIndicator.input,
+          animationConfig.gestureIndicator.output,
+        ),
+      }}
+      aria-hidden="true"
+      className={cx(
+        // Make sure to show the gesture indicator depending on whether it is touch screen device
+        styles.gestureIndicator,
+        isTouchDevice ? styles.touch : styles.scroll,
+      )}
```
Immo-Be: i feel like this is logic we would need for all scroll stories

src/scripts/components/stories/story/scroll-story/story-x-fires/splashscreen/splashscreen-x-fires.tsx
```diff
@@ -0,0 +1,106 @@
...
+  },
+  title: {
+    input: [0.5, 1],
+    output: ["-10vh", "-50vh"],
+  },
+};
+
+export type SplashAnimationConfig = typeof animationConfig;
+
+const GestureIndicator = () => {
```
Immo-Be: let's make this a separate component we can re-use

Immo-Be (approved):

Issue #2403, opened 2026-08-03
check progress indicator issues


Immo-Be:
- [x] check mobile
- [x] check scroll stories

annikabock:
<img width="1332" height="761" alt="Image" src="https://github.com/user-attachments/assets/e8a10f62-8cdb-4e9c-83c3-934b235f5a8e" />
still overlaying content here and there

Immo-Be:
ok, i see. This is the debug story, right?

PR #2424, opened 2026-08-04
fix(chapter-indicator): fix chapter indicator in scroll stories

Immo-Be:
for #2403 

- fixes the issue where for the scroll stories ("eei" and "xfires") the chapter indicator would not work correctly. This is because the ref to determine module elements was not handled correctly in the splashscreen

### DIFF
--- a/src/scripts/components/stories/story/blocks/splashscreen/splashscreen.tsx
+++ b/src/scripts/components/stories/story/blocks/splashscreen/splashscreen.tsx
@@ -1,6 +1,6 @@
-import { FunctionComponent, useEffect, useRef } from "react";
+import { FunctionComponent, RefObject, useEffect  } from "react";
 import { useDispatch } from "react-redux";
-import { motion, useTransform } from "motion/react";
+import { motion,  useTransform } from "motion/react";
 
 import { setFlyTo } from "../../../../../reducers/fly-to";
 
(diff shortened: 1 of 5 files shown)


src/scripts/components/stories/story/blocks/splashscreen/splashscreen.tsx
```diff
@@ -1,6 +1,6 @@
-import { FunctionComponent, useEffect, useRef } from "react";
+import { FunctionComponent, RefObject, useEffect  } from "react";
 import { useDispatch } from "react-redux";
-import { motion, useTransform } from "motion/react";
+import { motion,  useTransform } from "motion/react";
```
plumdumpling: Looks like your prettier isn't running. Can you please check?

src/scripts/components/stories/story/scroll-story/modules/base-scroll/module/scroll-module.tsx
```diff
@@ -41,16 +37,24 @@ const StickyContainer = ({
   );
 };
 
+/**
+ * We use this as a wrapper to for scroll module. It provides us with the current absolute and relative scroll position within the story
+ * This component may receive a refTarget from a parent component e.g. to add it to mooduleRefsMap which is used to track progress of story in the chapter-progress-indicator
```
plumdumpling:
```suggestion
 * We use this as a wrapper for the scroll module. It provides us with the current absolute and relative scroll position within the story
 * This component may receive a refTarget from a parent component e.g. to add it to moduleRefsMap which is used to track progress of story in the chapter-progress-indicator
```

src/scripts/types/story.ts
```diff
@@ -1,4 +1,4 @@
-import { ComponentProps, FunctionComponent } from "react";
+import { ComponentProps, FunctionComponent, Ref, RefObject } from "react";
```
plumdumpling:
```suggestion
import { ComponentProps, FunctionComponent, Ref } from "react";
```

plumdumpling (approved):

PR #2427, opened 2026-08-05
fix(image-scroll): correctly apply motion transform values

### LINKED ISSUE #2398: image scroll not working
some motion controlled properties are not updating correctly anymore. Works fine in firefox, does not work in safari and chrome. We need to fix this

Image

Immo-Be:
closes #2398

### DIFF
--- a/src/scripts/components/stories/story/blocks/image-gallery/modules/image-scroll/image-scroll-image/image-scroll-image.module.css
+++ b/src/scripts/components/stories/story/blocks/image-gallery/modules/image-scroll/image-scroll-image/image-scroll-image.module.css
@@ -1,6 +1,4 @@
 .scrollImageWrapper {
-  --custom-clip-path: inset(10% 10% 10% 10%);
-  --custom-button-opacity: 0;
   --container-bg: rgba(0, 0, 0, 0.9);
 }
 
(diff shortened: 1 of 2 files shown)


plumdumpling (commented):
Did you try using
`--custom-clip-path: isFullscreen ? "none" : clipPath`
instead of
`clipPath: isFullscreen ? "none" : clipPath`
and
`--custom-button-opacity: buttonOpacity`
instead of
`opacity: buttonOpacity`
in the style attribute of the elements?
This was my workaround for the opacity transition in the dimmer component of the xFires story. (I wanted to mention that in my PR, but I forgot.)

Immo-Be:
> Did you try using `--custom-clip-path: isFullscreen ? "none" : clipPath` instead of `clipPath: isFullscreen ? "none" : clipPath` and `--custom-button-opacity: buttonOpacity` instead of `opacity: buttonOpacity` in the style attribute of the elements? This was my workaround for the opacity transition in the dimmer component of the xFires story. (I wanted to mention that in my PR, but I forgot.)

That works, thanks :) so you experienced the same issue?

plumdumpling:
> That works, thanks :) so you experienced the same issue?

Yes

plumdumpling (approved):

PR #2428, opened 2026-08-05
Feat/video content in image module

### LINKED ISSUE #2386: Video content module
- [x] imageScrol
- [x] carousell

let's use the existing image module in the story mapper and just make it possible to play videos. is that an option or can't we detect what is an image and what is a video to display the video play button?

if that works brilliant. If not please add another tab for video module to the input sheet and the story mapper

We use the same technology that is implemented already

Image

Image

Immo-Be:
closes #2386


src/scripts/libs/is-video.ts
```diff
@@ -0,0 +1,5 @@
+const VIDEO_EXTENSIONS = [".mp4", ".webm", ".ogg", ".mov"];
+
+export const isVideo = (url?: string) => {
+  return VIDEO_EXTENSIONS.some((ext) => url?.toLowerCase().includes(ext));
```
plumdumpling, line 4: Just in case someone names a file something like `cover.movie.png`, `musikverein.oggelshausen.jpg` or similar. 🤷‍♀️  
```suggestion
  return VIDEO_EXTENSIONS.some((ext) => url?.toLowerCase().endsWith(ext));
```

src/scripts/components/stories/story/blocks/image-gallery/modules/shared/in-view-video/in-view-video.module.css
```diff
@@ -0,0 +1,37 @@
...
+
+  .playButton {
+    display: flex;
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    justify-content: center;
+    align-items: center;
+    gap: 10px;
+    padding: 8px 24px;
```
plumdumpling, line 17:
```suggestion
    gap: 0.625rem;
    padding: 0.5rem 1.5rem;
```

plumdumpling (approved):

PR #2443, opened 2026-08-11
Feat/scroll image sequence

Immo-Be:
for #2444 

- use a sequence of images rendered to a `canvas` to sync video playback to scroll

https://github.com/user-attachments/assets/631b45ee-ff05-4ab0-9f04-c9553087aa78

➡︎  Image sequences work better for scroll sync because each frame can be selected and drawn directly, giving better control at any scroll position. Video playback which i initially tried, relies on seeking and decoding around keyframes, which is optimized for continuous playback and often caused lag and stutter when matching to scroll progress interactively.

**Implemenentation:**

-  the implementation is inspired by a codepen on the gsap website: https://gsap.com/community/forums/topic/30947-smoothly-play-video-on-scrolltrigger/

-  for our purposes, I added a node script `extract-video-frames.js` (ai generated) that takes as input a video file (.mp4 etc.) and converts it to the sequence of images. I have used sensible defaults for, however, we might need to tweak them later. Furthermore, the script also auto-generates a `image-sequence.json` manifest which contains useful metadata to retrieve and pre-load sequence in the `ScrollImageSequence` component

- as for the output format, I decided to use `*.webp` because it is smaller, optimized for web and [widely supported](https://caniuse.com/?search=webp).

**How to add an ImageSequence to x-fires story:**
1. make sure you have `ffmpeg` installed 
2. execute  `extract-video-frames.js` script by passing input file and output directory (script generates files and manifest)
3. add the this to the module (should work with any module)
```
      "imageSequence": {
        "path": "assets/your-path-here"
      },
```

5. add a `ScrollImageSequence` component to the module root.
7. Optional: add a `progressRange` to the `animatConfig` (this maps the start and end to a specific point of the scrollProgress). E.g. [0.2, 0.8] would make the video start at 20% progress of the module and end by 80%. Defaults to [0,1].


plumdumpling (approved):
The last few frames seem to be all black. Can you please check and remove them if I'm right?

PR #2448, opened 2026-08-12
refactor(scroll-story): fetch image frames from cloud storage in dev

Immo-Be:
- fetch image sequence frames from cloud storage


storage/stories/story-x-fires/story-x-fires-config.json
```diff
@@ -169,6 +169,25 @@
       ],
       "lengthFactor": 14
     },
+    {
```
plumdumpling, line 217: Shouldn't be part of this PR?

storage/stories/story-x-fires/story-x-fires-en.json
```diff
@@ -37,6 +37,13 @@
         "fireRingIllustrationDescription": "Illustration showing burned and ground areas.",
         "legendLabelBurnedAreaThreshold": "Burned Area Threshold"
       }
+    },
```
plumdumpling, line 48: Same here?

storage/stories/story-x-fires/story-x-fires-config.json
Immo-Be: you're right. this was / is part of the pr for shot-9

plumdumpling (approved):

Issue #2388, opened 2026-07-16
Interactive graphs - Implementation Status


andreashelms:
[story45_graph_data.zip](https://github.com/user-attachments/files/31455289/story45_graph_data.zip)

[flatgraphs-csv.zip](https://github.com/user-attachments/files/31455354/flatgraphs-csv.zip)

[flatgraphs-csv-2.zip](https://github.com/user-attachments/files/31455373/flatgraphs-csv-2.zip)

Here is current status of charts implementation:

https://docs.google.com/spreadsheets/d/1yNNNlawieh44rk_Z5DtACfkMrqOqyQFdghjlB7pzGlI/

Note that some charts are already on develop branch but still based on mock data.
Update charts with real data before publishing.

annikabock:
[flatgraphs-csv 2.zip](https://github.com/user-attachments/files/31030438/flatgraphs-csv.2.zip)

these are the csv i currently have

andreashelms:
@annikabock Chart in story-21 (See https://github.com/ubilabs/esa-climate-from-space/pull/2487) shows different values that the original one (https://storage.googleapis.com/esa-cfs-versions/web/develop/index.html#/land/stories/story-21/1) because provided source data is different than the data displayed in the original chart. How should we proceed with that?

annikabock:
I left a comment in the overview sheet for PE

BastianGuenther:
Philip:
> They are not missing - they don’t exist. ie, the source data is not available to us.
>
> Going forward, we will try and provide source data for any new charts, but for the legacy content, some charts will have to remain as flat images.

PR #2487, opened 2026-08-25
feat(stories): add interactive vega chart to story-21

andreashelms:
for #2388

image


andreashelms:
According to Philip Eales no valid chart source data exists, so this can be closed.

Issue #2495, opened 2026-08-27
Review and optimise scroll-timing for desktop and mobile


annikabock:
lets do this at the very end - not too much scrolling, not too speedy layers coming in (shot 8 for desktop eg) please smooth this out.

Immo-Be:
@andreashelms could you check the current scrolling of the `x-fires` on android and give feedback? 🙏🏻

PR #2507, opened 2026-09-01
chore(x-fires): make scroll smoother, less "weighty"

Immo-Be:
for #2495


andreashelms:
Feels way smoother now 🔥

andreashelms (approved):

Immo-Be:
@andreashelms could you check again now? I felt like on my iphone i had to scroll too much still...

PR #2510, opened 2026-09-01
fix(ring-of-fire): ensure smooth scroll on ios

Immo-Be:


src/scripts/components/stories/story/scroll-story/story-x-fires/modules/burned-area/fire-ring/fire-ring.tsx
```diff
@@ -11,6 +11,7 @@ import {
 import { StoryXFiresModule } from "../../../../../../../../types/story";
 import { useModuleContent } from "../../../../../../../../providers/story/module-content/use-module-content";
 import { useScreenInfo } from "../../../../../../../../hooks/use-screen-info";
+import { isIos } from "../../../../../../../../libs/is-ios";
```
andreashelms: We already have lib functions `isIos16orLower()` and `isAndroid()`. Do we need this additional one?

andreashelms (approved):

PR #2532, opened 2026-09-09
fix(eei-story): layer loading delay on mobile

andreashelms:


src/scripts/hooks/use-story-layer.ts
```diff
@@ -1,20 +1,26 @@
...
+  const thunkDispatch = useThunkDispatch();
   const { mainId } = useSelector(selectedLayerIdsSelector);
 
-  const selectStoryLayer = useEffectEvent(() => {
-    dispatch(
-      setSelectedLayerIds({ layerId: initialLayerId, isPrimary: true }),
-    );
+  const initStoryLayer = useEffectEvent(() => {
+    // Fetch the layer details for all layers in the story to prevent loading delays.
+    storyLayerIds.forEach((layerId) => {
```
Immo-Be, line 19: how does this effect users with slow internet connection? There is no loading spinner so it would just show blank? We talked about pre-fetching but haven't implemented it yet, shouldn't we consider implications?

andreashelms: This is executed in background right when loading the story, before even showing any of those layers. This does not affect the behavior for stories other than x-fires / eei.

Immo-Be (approved):

PR #2553, opened 2026-09-15
chore(ci): add versioned web deploy cloud build for master

Immo-Be:
currently, there is no cloud build for the versioned build in storage/web/{version}/ which we use in the github release 
This adds a `cloudbuild.yaml` which will be triggered on push to `master` which will trigger the build


andreashelms:
> currently, there is no cloud build for the versioned build in storage/web/{version}/ which we use in the github release This adds a `cloudbuild.yaml` which will be triggered on push to `master` which will trigger the build

Can you remove this step from "Release New Version" documentation in readme then?

andreashelms (approved):

